PR #3, opened 2022-01-07
Initial Setup

jorge-ramirez-arredondo:
Started off as [create](https://github.com/BrightspaceUI/create) and modified to match closer to the setup [core](https://github.com/BrightspaceUI/core) has.

I recommend looking through the commits to get a better sense of the changes made from the base `create` project.


src/lang/ar.js
```diff
@@ -0,0 +1,5 @@
+/* eslint quotes: 0 */
+
+export default {
+	"hello": "مرحبا"
```
jorge-ramirez-arredondo, line 4: Should I remove all these `"hello"` lang terms?

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
dlockhart, line 6: You'll want to add these otherwise it'll try to upgrade Lit on you:

```
ignore:
      # Lit 2 upgrade
      - dependency-name: "@open-wc/testing"
        versions: ["3.x"]
      # Lit 2 upgrade
      - dependency-name: "lit-element"
        versions: ["3.x"]
      # Lit 2 upgrade
      - dependency-name: "lit-html"
        versions: ["2.x"]
```

package.json
```diff
@@ -0,0 +1,50 @@
...
+    "stylelint": "^14"
+  },
+  "files": [
+    "labs.serge.json",
+    "/src",
+    "!demo",
+    "!screenshots",
+    "!test",
+    "/src/components/demo",
+    "!/src/components/demo/demo"
```
dlockhart, line 41: I think you don't need these last 2 entries -- core has a `components/demo` component that we WANT to include (but not its demo) and labs won't have that.

src/lang/ar.js
dlockhart: Hmm, yeah if your initial thing isn't going to have any lang terms maybe we leave adding localization/Serge as an exercise for the next person?

dlockhart (approved):
This looks excellent to me.

package.json
jorge-ramirez-arredondo: Ah okay, that makes more sense now.

jorge-ramirez-arredondo:
@dlockhart Both checks are failing because they can't find any test files. Seeing as this doesn't have any tests yet, should we merge it as is, or should we change it so it doesn't fail? I suspect this error exists so that tests don't get missed if there are any, so I'm not sure if we should change anything.

dlockhart:
I think for now it's OK to merge this as-is, assuming we're planning to add both types of tests in the very near future.

PR #17, opened 2022-06-02
feat: add support for custom controllers within the ComputedValues controller

jorge-ramirez-arredondo:
This change will allow custom controllers to be used within the ComputedValues array of controllers. This is needed in order to be able to control the order of execution of compute methods between custom controllers and any other controllers within the ComputedValues controller.


dlockhart (approved):

svanherk:
I've disabled the visual-diff workflow until we have some actual visual-diff tests lol.  It's not a required check so won't block merging.

PR #28, opened 2022-09-30
Folder and Readme for Insights

NicholasHallman:
We have a few new shared components in the work so we are setting aside a spot for our work. Expecting to add a
 - telementry mixin
 - fullscreen component mixin
 - chart component with extensions

and hoping to migrate in

 - router
 - accessible toaster
 - immersive nav bar
 - card cover

### DIFF
--- a/src/insights/README.md
+++ b/src/insights/README.md
@@ -1,0 +1,5 @@
+# Insights Components
+
+Components, Controllers and Mixins owned by team Nightingale.
+
+@Brightspace/draco-vis


dlockhart (commented):
I'd prefer if we didn't organize things by team, since that makes components less discoverable. Can we just have `src/components/*`, `src/controllers/*`, `src/mixins/*`, etc.?

NicholasHallman:
Sounds good I'll close this

PR #31, opened 2022-10-27
Migrate to node 16

oelhanafey:
[US144541](https://rally1.rallydev.com/#/?detail=/userstory/662806005029&fdp=true)

This PR migrates the repo to node 16.

Our visual-diff library has a peer dependency on Puppeteer. In Node >= 16, peer dependencies are automatically installed, which would allow our visual-diff action to just install the library and be done with it. Instead, we need to coordinate upgrades each time Puppeteer makes a new release for repos that are running Node < 16.


dlockhart (approved):
That was easy!

PR #32, opened 2022-11-14
Init Drawer component in labs

KearseTrevor:
Letting drawer finally see some daylight rather than sitting locally in on my dust shelf.
Drawer is a _modal that is anchored to the edge of the viewport or parent container_

The readme.md placed within the drawer directory explains more than I will in this PR description
- Essentially, there's still a lot to do on this component but I feel like it's a waste to just let it sit forgotten on my desktop.
- A couple of these features are more or less critical though, IMO so without them I didn't even consider this component ready for consumption on a labs level so I'm not exporting it via package.json at this moment in time

Took inspiration from the following sources for drawer:
[Shoelace](https://shoelace.style/components/drawer)
[Mantine](https://mantine.dev/core/drawer/)
[Material-UI](https://mui.com/material-ui/react-drawer/#main-content)

Also did some housekeeping in the repo:
- Updated .eslint.json for demo console logs & lang file quotes
- Added a new mixin for `localizeLabsElement` which is a copy paste of `localizeCoreElement` for labs repo

Should note that this component has a decent amount in common with `dialog.js` from core - so anyone familiar with that code would probably have the most input on this component?

Demo screenshot
image

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,3 +1,19 @@
 {
-  "extends": "brightspace/lit-config"
+  "extends": "brightspace/lit-config",
+  "overrides": [{
+    "files": [
+      "**/demo/**/*.html"
+    ],
+    "rules": {
+      "no-console": 0
+    }
+  },
+  {
+    "files": [
+      "**/lang/*.js"
+    ],
+    "rules": {
+      "quotes": 0
+    }
+  }]
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,19 +1,3 @@
 {
-  "extends": "brightspace/lit-config",
-  "overrides": [{
-    "files": [
-      "**/demo/**/*.html"
-    ],
-    "rules": {
-      "no-console": 0
-    }
-  },
-  {
-    "files": [
-      "**/lang/*.js"
-    ],
-    "rules": {
-      "quotes": 0
-    }
-  }]
+  "extends": "brightspace/lit-config"
 }
--- a/demo/components/drawer/index.html
+++ b/demo/components/drawer/index.html
@@ -8,12 +8,12 @@
             import '@brightspace-ui/core/components/button/button.js';
             import '../../../src/components/drawer/drawer.js';
         </script>
-        <title>Drawer</title>
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <meta charset="UTF-8">
     </head>
-    <body>
+    <body unresolved>
         <d2l-demo-page page-title="Drawer">
+
             <h2>Drawer Positions</h2>
             <d2l-demo-snippet>
                 <template>
@@ -89,17 +89,18 @@
                     </script>
                 </template>
             </d2l-demo-snippet>
+
 			<h2>Drawer Sizes</h2>
             <d2l-demo-snippet>
                 <template>
-                    <d2l-button id="open-xs">xs</d2l-button>
-                    <d2l-button id="open-sm">sm</d2l-button>
-                    <d2l-button id="open-md">md</d2l-button>
-                    <d2l-button id="open-lg">lg</d2l-button>
-					<d2l-button id="open-xl">xl</d2l-button>
+                    <d2l-button id="open-xs">xsmall</d2l-button>
+                    <d2l-button id="open-sm">small</d2l-button>
+                    <d2l-button id="open-md">medium</d2l-button>
+                    <d2l-button id="open-lg">large</d2l-button>
+					<d2l-button id="open-xl">xlarge</d2l-button>
 					<d2l-button id="open-full">full</d2l-button>
-					<d2l-input-text id="custom-size-input" label="Custom Size Input" placeholder="% or px string"></d2l-input-text>
-					<d2l-button id="open-custom">Custom Size</d2l-button>
+					<d2l-button id="open-400">400px</d2l-button>
+					<d2l-button id="open-25-percent">25%</d2l-button>
                     <d2l-labs-drawer id="size-demo" label="Foobar">
                         <div>
                             <d2l-input-text id="nameId" label="Name" autofocus></d2l-input-text>
@@ -115,37 +116,41 @@
                     </d2l-labs-drawer>
                     <script>
 						document.querySelector('#open-xs').addEventListener('click', () => {
-							document.querySelector('#size-demo').setAttribute('size', 'xs');
+							document.querySelector('#size-demo').setAttribute('size', 'xsmall');
 							document.querySelector('#size-demo').show();
 						});
 						document.querySelector('#open-sm').addEventListener('click', () => {
-							document.querySelector('#size-demo').setAttribute('size', 'sm');
+							document.querySelector('#size-demo').setAttribute('size', 'small');
 							document.querySelector('#size-demo').show();
 						});
 						document.querySelector('#open-md').addEventListener('click', () => {
-							document.querySelector('#size-demo').setAttribute('size', 'md');
+							document.querySelector('#size-demo').setAttribute('size', 'medium');
 							document.querySelector('#size-demo').show();
 						});
 						document.querySelector('#open-lg').addEventListener('click', () => {
-							document.querySelector('#size-demo').setAttribute('size', 'lg');
+							document.querySelector('#size-demo').setAttribute('size', 'large');
 							document.querySelector('#size-demo').show();
 						});
 						document.querySelector('#open-xl').addEventListener('click', () => {
-							document.querySelector('#size-demo').setAttribute('size', 'xl');
+							document.querySelector('#size-demo').setAttribute('size', 'xlarge');
 							document.querySelector('#size-demo').show();
 						});
 						document.querySelector('#open-full').addEventListener('click', () => {
 							document.querySelector('#size-demo').setAttribute('size', 'full');
 							document.querySelector('#size-demo').show();
 						});
-						document.querySelector('#open-custom').addEventListener('click', () => {
-							const customSize = document.querySelector('#custom-size-input').value;
-							document.querySelector('#size-demo').setAttribute('size', customSize);
+						document.querySelector('#open-400').addEventListener('click', () => {
+							document.querySelector('#size-demo').setAttribute('size', '400px');
+							document.querySelector('#size-demo').show();
+						});
+						document.querySelector('#open-25-percent').addEventListener('click', () => {
+							document.querySelector('#size-demo').setAttribute('size', '25%');
 							document.querySelector('#size-demo').show();
 						});
                     </script>
                 </template>
             </d2l-demo-snippet>
+
 			<h2>Prevent Closing</h2>
 			<d2l-demo-snippet>
 				<template>
--- a/demo/components/drawer/index.html
+++ b/demo/components/drawer/index.html
@@ -1,0 +1,200 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <link rel="stylesheet" href=/node_modules/@brightspace-ui/core/components/demo/styles.css type="text/css">
+        <script type="module">
+            import '@brightspace-ui/core/components/demo/demo-page.js';
+            import '@brightspace-ui/core/components/inputs/input-text.js';
+            import '@brightspace-ui/core/components/button/button.js';
+            import '../../../src/components/drawer/drawer.js';
+        </script>
+        <title>Drawer</title>
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <meta charset="UTF-8">
+    </head>
+    <body>
+        <d2l-demo-page page-title="Drawer">
+            <h2>Drawer Positions</h2>
+            <d2l-demo-snippet>
+                <template>
+                    <d2l-button id="open-right">right</d2l-button>
+                    <d2l-button id="open-left">left</d2l-button>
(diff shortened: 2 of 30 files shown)


src/components/drawer/drawer-mixin.js
```diff
@@ -0,0 +1,509 @@
...
+			_overflowTop: { type: Boolean },
+			_parentDrawer: { type: Object },
+			_scroll: { type: Boolean },
+			_state: { type: String, reflect: true },
+		};
+	}
+
+	constructor() {
+		super();
+		this.label = 'Drawer';
```
dbatiste, line 53: Is this intended to be hard-coded?

src/components/drawer/drawer-mixin.js
```diff
@@ -0,0 +1,509 @@
...
+		} else {
+			this._close();
+		}
+	}
+
+	close() {
+		if (!this.open) return;
+		this.open = false;
+		return new Promise(resolve => {
+			setTimeout(async() => {
```
dbatiste, line 83: What's the reason for wrapping these in `setTimeout` calls?

src/components/drawer/drawer-mixin.js
```diff
@@ -0,0 +1,509 @@
...
+		});
+	}
+
+	show() {
+		if (this.open) return;
+		this.open = true;
+		return new Promise((resolve) => {
+			const onClose = function(e) {
+				if (e.target !== this) return; // ignore if bubbling from child drawer
+				this.removeEventListener('d2l-drawer-close', onClose);
```
dbatiste, line 105: It is possible to avoid having to call `removeEventListener` by specifying the `once` option when calling `addEventListener`:
```javascript
this.addEventListener('d2l-drawer-close', onClose, { once: true });
```
I think it should even be possible to avoid the explicit binding call by using an arrow function. Something like this:
```javascript
this.addEventListener('d2l-drawer-close', e => {
    if (e.target !== this) return; // ignore if bubbling from child drawer
    resolve(e.detail.action);
}, { once: true });
```

src/components/drawer/drawer-mixin.js
```diff
@@ -0,0 +1,509 @@
...
+                    @d2l-focus-trap-enter=${this._handleFocusTrapEnter}
+                    ?trap=${this.open && this.trapFocus}>
+                    ${inner}
+                </d2l-focus-trap>
+            </div>
+            <d2l-backdrop
+                for-target=${this._drawerId}
+                ?shown=${this._state === 'showing'}
+                @click=${this._handleBackdropClick}
+                tabindex="-1">
```
dbatiste, line 487: What's the `tabindex="-1"` for on the backdrop?

src/components/drawer/drawer.js
```diff
@@ -0,0 +1,162 @@
...
+                <d2l-button-icon
+                    id="drawer-default-close"
+                    icon="tier1:close-small"
+                    text="${this.localize('components.drawer.close')}"
+                    @click=${this._abort}
+                    tabindex=100>
+                </d2l-button-icon>`
+			: nothing;
+	}
+	get content() {
```
dbatiste, line 83: Is it intended for `content` and `header` to be public properties? Normally if we want to break up the rendering we will create internal helper methods, `_renderHeader() { ... }`.

test/components/drawer/drawer.visual-diff.js
```diff
@@ -0,0 +1,28 @@
...
+		await page.bringToFront();
+	});
+
+	beforeEach(async() => {
+		await visualDiff.resetFocus(page);
+	});
+
+	after(async() => await browser.close());
+
+	it('passes visual-diff comparison', async function() {
```
dbatiste, line 23: Would definitely be good to eventually have diffs for the drawer in different configurations, states. 😄

dbatiste (commented):
Hey Trevor, per our discussion offline, here are a few things that jump out at me on first look.

.eslintrc.json
```diff
@@ -1,3 +1,19 @@
 {
-  "extends": "brightspace/lit-config"
+  "extends": "brightspace/lit-config",
+  "overrides": [{
+    "files": [
+      "**/demo/**/*.html"
+    ],
+    "rules": {
+      "no-console": 0
```
dlockhart, line 8: Can we move this override into an `.eslintrc.json` inside the `demo` directory?

.eslintrc.json
```diff
@@ -1,3 +1,19 @@
...
+    }
+  },
+  {
+    "files": [
+      "**/lang/*.js"
```
dlockhart, line 13: Similarly can we move this into an `.eslintrc.json` in `src`?

demo/components/drawer/index.html
```diff
@@ -0,0 +1,200 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <link rel="stylesheet" href=/node_modules/@brightspace-ui/core/components/demo/styles.css type="text/css">
```
dlockhart, line 4:
```suggestion
        <link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
```

helpers/localizeLabsElement.js
```diff
@@ -0,0 +1,11 @@
...
+
+export const LocalizeLabsElement = superclass => class extends LocalizeDynamicMixin(superclass) {
+
+	static get localizeConfig() {
+		return {
+			importFunc: async lang => (await import(`../src/lang/${lang}.js`)).default
+		};
+	}
+
+};
```
dlockhart, line 11: This is probably fine for now. As more and more components get added here, it may make sense to split their localization out to avoid having the user download a bunch of translations for components that likely aren't rendered on the page.

src/components/drawer/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Usage
+
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/button/button.js';
+  import '@brightspace-ui-labs/drawer/drawer.js';
+
+  document.querySelector('#open-demo').addEventListener('click', () => {
+    document.querySelector('#drawer-demo').show() = true;
```
dlockhart, line 24: Should this be:
```suggestion
    document.querySelector('#drawer-demo').show();
```

src/components/drawer/drawer-mixin.js
```diff
@@ -0,0 +1,509 @@
...
+			_parentDrawer: { type: Object },
+			_scroll: { type: Boolean },
+			_state: { type: String, reflect: true },
+		};
+	}
+
+	constructor() {
+		super();
+		this.label = 'Drawer';
+		this.open = false;
```
dlockhart, line 54: Can this be called `opened` just so that it matches dialog's API?

Also, instead of having `show()` and `hide()` as public methods, can setting the attribute/property be the thing that triggers showing and hiding? The [motivation for that approach is here](https://daylight.d2l.dev/developing/best-practices/methods/#component-architecture).

The component is already working that way, so the change would be to make `_show()` and `_hide()` internal I think and update the docs to tell people to just get the element and call `elem.opened = true`.

src/components/drawer/README.md
```diff
@@ -0,0 +1,90 @@
...
+| `closeButtonLabel` | Close button aria-label | `string` | `'Close'` |
+| `label` | Drawer's label as displayed via aria/screenreader & title of header | `string` | `'Drawer'` |
+
+**Events:**
+
+| Name | Description |
+| --- | ---- |
+| `d2l-drawer-closed` | Emitted when drawer has closed and animations are complete |
+| `d2l-drawer-opened` | Emitted when drawer has opened and animations are complete |
+| `d2l-drawer-request-close` | Emitted when drawer close is requested - provides opportunity to prevent closure |
```
dlockhart, line 77: Can these events be prefixed as `d2l-labs-drawer-*`? That way if we want to change them if this component moves beyond Labs, it's not a breaking change.

src/components/drawer/drawer-mixin.js
```diff
@@ -0,0 +1,509 @@
...
+		return new Promise(resolve => {
+			setTimeout(async() => {
+				await this._close();
+				resolve();
+			}, 0);
+		});
+	}
+
+	resize() {
```
dlockhart, line 90: Why does this need to be a public method?

src/components/drawer/drawer-mixin.js
KearseTrevor: stemmed from a copy and paste misnomer - `resize` on dialog-mixin.js in core required it for timing purposes in Firefox https://github.com/BrightspaceUI/core/pull/2233.
Seems related to use of native dialog component though and therefore should have no influence on drawer - will remove.

src/components/drawer/drawer.js
```diff
@@ -0,0 +1,162 @@
...
+
+const mediaQueryList = window.matchMedia('(max-width: 615px), (max-height: 420px) and (max-width: 900px)');
+
+/**
+ * Modal element that is anchored to one side of the viewport or parent element.
+ * @fires d2l-drawer-before-close - Dispatched with bound action value before the dialog is closed for any reason, providing an opportunity to prevent the closing.
+ * @slot - Default slot for content inside drawer
+ * @slot footer - Slot for footer content such as workflow buttons. Apply attribute `drawer-close` to workflow buttons to have them automatically close the drawer with the action's bound `value`.
+ */
+class Drawer extends LocalizeLabsElement(DrawerMixin(LitElement)) {
```
dlockhart, line 21: Is the mixin even needed? I know this was based off dialog, but it's using a mixin to share behaviour across several components.

src/lang/ar.js
```diff
@@ -1,4 +1,3 @@
-/* eslint quotes: 0 */
```
dlockhart, line 1: I have a feeling Serge is going to re-add this comment back when the translations come back...

eKoopmans:
Looks like the Daves have this well in hand, but I do have a question:

It seems we already have existing drawer functionality, e.g. in the [dropdown "mobile-tray"](https://daylight.d2l.dev/components/dropdown/#d2l-dropdown-content-properties). What's the relationship between that existing functionality and this component?

.eslintrc.json
KearseTrevor: Sure thing, was following pattern I saw in core library for rule exceptions - I have no issue moving spot though. Suggest updating pattern in core as well if this is the preference.

dlockhart: Thanks. Yeah I wish core was set up the way this repo is -- with `demo/` and `test/` at the root instead of nested inside each component directory. This setup allows the `eslint` configs for demo/test/src to exist in just 1 place instead of copied into each sub-directory.

dbatiste:
>It seems we already have existing drawer functionality, e.g. in the [dropdown "mobile-tray"](https://daylight.d2l.dev/components/dropdown/#d2l-dropdown-content-properties). What's the relationship between that existing functionality and this component?

That's a good question. They are really similar eh? Maybe dropdown currently doesn't offer the degree of freedom in terms of positioning, API, and styling in order to do this, but in theory it probably could. Semantically, I think they are the same thing, though I also get dialog vibes. Of course, from a naming perspective, this isn't dropdown-y at all. If I were starting from scratch I would be looking at some type of base popup component or behaviour (and thinking in advance to perhaps even using the new `popup` API), and then ways to apply different positioning/sizing logic to it, along with the ability to customize the style/decorate with pointer.

src/components/drawer/drawer.js
KearseTrevor: I was at the stage where I wasn't sure what the future of this component would hold and didn't know if there would ever be fracturing along use case lines - so I figured I would keep as a mixin for now and fold it in if it wasn't needed

svanherk:
> > It seems we already have existing drawer functionality, e.g. in the [dropdown "mobile-tray"](https://daylight.d2l.dev/components/dropdown/#d2l-dropdown-content-properties). What's the relationship between that existing functionality and this component?
> 
> That's a good question. They are really similar eh? Maybe dropdown currently doesn't offer the degree of freedom in terms of positioning, API, and styling in order to do this, but in theory it probably could. Semantically, I think they are the same thing, though I also get dialog vibes. Of course, from a naming perspective, this isn't dropdown-y at all. If I were starting from scratch I would be looking at some type of base popup component or behaviour (and thinking in advance to perhaps even use the new `popup` API), and then ways to apply different positioning/sizing logic to it, along with the ability to customize the style/decorate with pointer. Then really thin wrappers on that, for dropdown, drawer, etc.

There is also the mobile menu tray in the LMS that dropdown's is based on.  I think the plan was to update it to use `dropdown` instead, but until then it's currently another iteration.

KearseTrevor:
Promising to return to this before it gets too stale - I've got some cycles soon that I'll be using to return to this PR.

KearseTrevor:
So apparently my definition of "before it gets too stale" is a little loose.

I'll be taking a look at this tomorrow - *most likely* it will be killed via death-by-stale but I'll be doing some exploration in the Daylight catalogue for what exists today and see if the core desire that spurred this component has been addressed or not.

If the need hasn't been met I'll sidebar with Dave B and Jeff to discuss work that I could do on the side for this as I would still like to kick this thing across the metaphorical finish line one day - if only for hubris sake
  - Will probably just relaunch the project from scratch so this PR would be closed in this case anyway
  - Suppose that means I'm just closing this PR NOW will follow up with additional action items at another point in time

Rest in peace v0.1 of `drawer` you were gone before we even knew you

PR #43, opened 2023-06-06
Update package-lock.json

jorge-ramirez-arredondo:
update-package-lock action is failing because it seems the committed package-lock.json file wasn't correct? I thought I did this right when I first committed it, but maybe I made a mistake, so I ran npm install again to see if that does it.

image
https://github.com/BrightspaceUI/labs/actions/runs/5193348328/jobs/9363776901

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9679,16 +9679,18 @@
       }
     },
     "node_modules/typescript": {
-      "version": "3.9.10",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-3.9.10.tgz",
-      "integrity": "sha512-w6fIxVE/H1PkLKcCPsFqKE7Kv7QUwhU8qQY2MueZXWx5cPZdwFupLgKK3vntcK98BtNHZtAF4LA/yl2a7k8R6Q==",
+      "version": "5.1.3",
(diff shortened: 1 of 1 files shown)


dlockhart:
This is a known problem with `lit-analyzer` -- in other repos I've just [removed the analyzer entirely](https://github.com/BrightspaceHypermediaComponents/users/pull/156) since it doesn't provide much value anymore.

jorge-ramirez-arredondo:
@dlockhart I noticed in core you [still have lit-analyzer](https://github.com/BrightspaceUI/core/blob/main/package.json#L19). How did you solve this there? Did you just [include typescript 5](https://github.com/BrightspaceUI/core/blob/main/package.json#L70)?

dlockhart:
Yeah we did, but only because core is relying on `lit-analyzer` to generate the `custom-elemens.json` metadata file that the Daylight Site uses. Otherwise I'd remove it.

dlockhart:
Haha, maybe! Ok that makes me want to go and rip it out... brb.

jorge-ramirez-arredondo:
Okay, this PR should no longer be needed now that lit-analyzer has been removed as part of https://github.com/BrightspaceUI/labs/pull/44

PR #98, opened 2023-08-11
feat: move opt-in flyout component into labs

dlockhart:
This moves the opt-in-flyout from [its repo](https://github.com/BrightspaceUILabs/opt-in-flyout) into the Labs monorepo (here). Once BSI (the only consumer) is switched over, I'll Graveyard the other repo.

The old repo doesn't have vdiff tests, but after this part is in I'm going to add some. I think that will allow me to delete most of the "unit" tests, which are doing a lot of assertions about the presence/absence of particular elements.

### DIFF
--- a/demo/components/opt-in-flyout/index.html
+++ b/demo/components/opt-in-flyout/index.html
@@ -1,0 +1,94 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
+		<script type="module">
+			import '@brightspace-ui/core/components/demo/demo-page.js';
+			import '../../../src/components/opt-in-flyout/opt-in-flyout.js';
+			import '../../../src/components/opt-in-flyout/opt-out-flyout.js';
+			import '../../../src/components/opt-in-flyout/opt-out-reason.js';
+		</script>
+		<title>opt-in-flyout demo</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta charset="UTF-8">
+		<style>
+			.flyout-wrapper {
+				height: 450px;
+				position: relative;
+				width: 100%;
+			}
+		</style>
+	</head>
+	<body>
+		<d2l-demo-page page-title="Opt-in Flyout">
+
+			<d2l-demo-snippet no-padding>
+				<div class="flyout-wrapper">
+					<d2l-opt-in-flyout
+						open
+						title="Flyout Demo Opt-in"
+						short-description="This is a <b>short</b> description" long-description="This is a <b>long</b> description"
+						tab-position="right"
+						tutorial-link="https://www.example.com#tutorial"
+						help-docs-link="https://www.example.com#documentation">
+					</d2l-opt-in-flyout>
+				</div>
+			</d2l-demo-snippet>
+
+			<d2l-demo-snippet no-padding>
+				<div class="flyout-wrapper">
+					<d2l-opt-out-flyout
+						open
+						title="Flyout Demo Opt-out"
+						short-description="This is a <b>short</b> description"
+						long-description="This is a <b>long</b> description"
+						tab-position="right"
+						tutorial-link="https://www.example.com#tutorial"
+						help-docs-link="https://www.example.com#documentation">
+					</d2l-opt-out-flyout>
+				</div>
+			</d2l-demo-snippet>
+
+			<d2l-demo-snippet no-padding>
+				<div class="flyout-wrapper">
+					<d2l-opt-out-flyout
+						open
+						title="Flyout Demo Opt-out"
+						short-description="This is a <b>short</b> description"
+						long-description="This is a <b>long</b> description"
+						tab-position="right"
+						tutorial-link="https://www.example.com#tutorial"
+						help-docs-link="https://www.example.com#documentation">
+						<d2l-opt-out-reason key="test-1" text="Test Option 1"></d2l-opt-out-reason>
+						<d2l-opt-out-reason key="test-2" text="Test Option 2"></d2l-opt-out-reason>
+					</d2l-opt-out-flyout>
+				</div>
+			</d2l-demo-snippet>
+
+			<d2l-demo-snippet no-padding>
+				<div class="flyout-wrapper">
+					<d2l-opt-out-flyout
+						hide-reason
+						hide-feedback
+						open
+						title="Flyout Demo Opt-out - dialog disabled"
+						short-description="This is a <b>short</b> description"
+						long-description="This is a <b>long</b> description"
+						tab-position="right"
+						tutorial-link="https://www.example.com#tutorial"
+						help-docs-link="https://www.example.com#documentation">
+					</d2l-opt-out-flyout>
+				</div>
+			</d2l-demo-snippet>
+
+			<script type="module">
+				[...document.querySelectorAll('d2l-opt-in-flyout, d2l-opt-out-flyout')].forEach(elem => {
+					['flyout-opened', 'flyout-closed', 'opt-in', 'opt-out'].forEach(eventName => {
+						// eslint-disable-next-line no-console
+						elem.addEventListener(eventName, (event) => console.log(eventName, JSON.stringify(event.detail)));
+					});
+				});
+			</script>
+		</d2l-demo-page>
+	</body>
+</html>
--- a/demo/components/opt-in-flyout/index.html
+++ b/demo/components/opt-in-flyout/index.html
@@ -24,20 +24,20 @@
 
 			<d2l-demo-snippet no-padding>
 				<div class="flyout-wrapper">
-					<d2l-opt-in-flyout
+					<d2l-labs-opt-in-flyout
 						open
 						title="Flyout Demo Opt-in"
 						short-description="This is a <b>short</b> description" long-description="This is a <b>long</b> description"
 						tab-position="right"
 						tutorial-link="https://www.example.com#tutorial"
 						help-docs-link="https://www.example.com#documentation">
-					</d2l-opt-in-flyout>
+					</d2l-labs-opt-in-flyout>
 				</div>
 			</d2l-demo-snippet>
 
 			<d2l-demo-snippet no-padding>
 				<div class="flyout-wrapper">
-					<d2l-opt-out-flyout
+					<d2l-labs-opt-out-flyout
 						open
 						title="Flyout Demo Opt-out"
 						short-description="This is a <b>short</b> description"
@@ -45,13 +45,13 @@
 						tab-position="right"
 						tutorial-link="https://www.example.com#tutorial"
 						help-docs-link="https://www.example.com#documentation">
-					</d2l-opt-out-flyout>
+					</d2l-labs-opt-out-flyout>
 				</div>
 			</d2l-demo-snippet>
 
 			<d2l-demo-snippet no-padding>
 				<div class="flyout-wrapper">
-					<d2l-opt-out-flyout
+					<d2l-labs-opt-out-flyout
 						open
 						title="Flyout Demo Opt-out"
 						short-description="This is a <b>short</b> description"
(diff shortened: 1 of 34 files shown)


demo/components/opt-in-flyout/index.html
```diff
@@ -0,0 +1,94 @@
...
+			.flyout-wrapper {
+				height: 450px;
+				position: relative;
+				width: 100%;
+			}
+		</style>
+	</head>
+	<body>
+		<d2l-demo-page page-title="Opt-in Flyout">
+
```
dlockhart, line 24: The existing demos had their own language switcher and a more complex "log" for events. I've simplified this and just log to the console.

package.json
```diff
@@ -4,6 +4,9 @@
   "type": "module",
   "repository": "https://github.com/BrightspaceUI/labs.git",
   "exports": {
+    "./components/opt-in-flyout.js": "./src/components/opt-in-flyout/opt-in-flyout.js",
+    "./components/opt-out-flyout.js": "./src/components/opt-in-flyout/opt-out-flyout.js",
+    "./components/opt-out-reason.js": "./src/components/opt-in-flyout/opt-out-reason.js",
```
dlockhart, line 9: Following the pattern for controllers -- I think this is nice actually as we get the benefit of the directories for code organization but when consumers are importing there's less repetition.

test/components/opt-in-flyout/utilities.js
```diff
@@ -0,0 +1,21 @@
...
+	}
+
+	const array = [];
+	for (let i = 0; i < collection.length; i++) {
+		array.push(collection[i]);
+	}
+	return array;
+};
+
+export default {
```
dlockhart, line 13: I'm going to revisit this when I add the vdiff tests. At the very least `toArray` can be replaced by the spread operator.

dlockhart:
Actually I'm going to rename these to be `<d2l-labs-*>` while I'm doing this -- will update this PR.

dbatiste:
> Actually I'm going to rename these to be `<d2l-labs-*>` while I'm doing this -- will update this PR.

👍 . Yeah I think that's a good call.

demo/components/opt-in-flyout/index.html
```diff
@@ -0,0 +1,94 @@
...
+<html lang="en">
+	<head>
+		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
+		<script type="module">
+			import '@brightspace-ui/core/components/demo/demo-page.js';
+			import '../../../src/components/opt-in-flyout/opt-in-flyout.js';
+			import '../../../src/components/opt-in-flyout/opt-out-flyout.js';
+			import '../../../src/components/opt-in-flyout/opt-out-reason.js';
+		</script>
+		<title>opt-in-flyout demo</title>
```
dbatiste, line 11: Note the `d2l-demo-page` will create this element automatically using the provided `page-title`. I think this may even result in duplicate `<title>` elements.

demo/components/opt-in-flyout/index.html
```diff
@@ -0,0 +1,94 @@
...
+		<meta charset="UTF-8">
+		<style>
+			.flyout-wrapper {
+				height: 450px;
+				position: relative;
+				width: 100%;
+			}
+		</style>
+	</head>
+	<body>
```
dbatiste, line 22: Do you want to include `unresolved` here. The `.../demo/styles.css` files contains the style to set... 
```css
body[unresolved] {
	opacity: 0;
}
```
When I was putting together my AI image description demo page I did notice I was seeing some FOUC without it. I think mostly resulting from the delay loading the font.

dlockhart: Sure -- I'll add that to the create template as well.

demo/components/opt-in-flyout/index.html
dlockhart: Thanks -- updated.

demo/components/opt-in-flyout/index.html
```diff
@@ -0,0 +1,94 @@
...
+						tab-position="right"
+						tutorial-link="https://www.example.com#tutorial"
+						help-docs-link="https://www.example.com#documentation">
+					</d2l-labs-opt-out-flyout>
+				</div>
+			</d2l-demo-snippet>
+
+			<script type="module">
+				[...document.querySelectorAll('d2l-labs-opt-in-flyout, d2l-labs-opt-out-flyout')].forEach(elem => {
+					['flyout-opened', 'flyout-closed', 'opt-in', 'opt-out'].forEach(eventName => {
```
dbatiste, line 85: If the demo snippet HTML isn't wrapped in a `<template>` then the formatted code that is displayed will contain any properties that are reflected as attributes which might not actually be part of the public API. This can be somewhat misleading to anyone looking at the demo code snippet so wrapping in a `<template>` is generally preferred. 

However in doing so, the rendering of the snippet elements becomes delayed due to the demo-snippet stamping and Lit's async rendering. This can result in JavaScript errors trying to query for the elements immediately. Three solutions I've used for that - 1) wrap the demo event code in a `setTimeout(...)` so that it's similarly delayed, or 2) wire up to `document.body` instead if the events bubble, which simplifies the script code slightly more; or 3) move the `<script>` containing the event wire-up into the `<template>` if we want to show that to the developer, though that means repeating it for each snippet and assigning ids so it can identify the correct element to wire up to.

So long story short, if we want this to be an example for other labs to copy, and we want them to use the `<template>` approach, then this script code would need to be tweaked slightly.

dlockhart: Interesting. I'll update the create template with this as well. I've updated it to wrap in a `<template>` and this code is actually still working... so... win?

dbatiste: Cool. Yeah so win!

I did a quick test with reflecting an internal property and it wasn't showing up in the formatted code, so maybe that got resolved as a result of some other changes to `d2l-demo-snippet` over the years. However, I did confirm that the constructor runs twice when `template` isn't used as a result of the way that `d2l-demo-snippet` grabs the content, clone and appends it internally, formats, etc., whereas the constructor only runs once when a `template` is used. So using `template` still seems preferred.

dlockhart: Gotcha. Now that vdiff is using `@web/test-runner`, it'll be interesting to see if we can update the demo snippet such that you could pass it a fixture that's defined elsewhere as an HTML literal, which would allow us to reuse these between demos and tests.

test/components/opt-in-flyout/opt-in-flyout.test.js
```diff
@@ -0,0 +1,183 @@
...
+
+		it('should contain leave it off button', () => {
+			const button = innerFlyout.shadowRoot.querySelector('d2l-button:not([primary])');
+			expect(button.textContent).to.equal('Leave it off');
+			expect(button).to.exist;
+		});
+
+		it('should fire opt-in event when enable button clicked', async() => {
+			const button = innerFlyout.shadowRoot.querySelector('d2l-button[primary]');
+			clickElem(button);
```
dbatiste, line 136: I'm guessing that `clickElem` has a `setTimeout` or rAF baked into it somewhere to ensure it runs after the event wire-up, ya?

dlockhart: Yeah exactly -- we've done a bunch of testing with these and confirmed that all the interaction events have sufficient timeouts. Behind the scenes, this is communicating via web-socket back to the server, which then instructions the browser to perform the task, which then communicates back.

demo/components/opt-in-flyout/index.html
svanherk: One potential negative effect of always using `template` and having the delayed rendering - for the `SkeletonMixin` stuff, it "fixed" the crazy import order problem we were seeing, but doesn't _really_ fix it since actually consumers aren't using a `template`.  Something we might just need to keep in mind, in theory tests would (hopefully) highlight the issue

dbatiste (approved):

test/components/opt-in-flyout/opt-in-flyout.test.js
dbatiste: On one hand, convenient. On the other, feels weird... I'm always going to be thinking to myself that all the clicking code will run before the event wire-up. I do worry a little about confusion if people aren't sure when they need to explicitly delay (using `setTimeout` or getting the promise and returning it) vs just trusting the helper to be delayed. I suppose if they are always using this helper or a helper that works in the same way, then there shouldn't be any need to write it differently, and this no source of confusion. 🤷‍♂️

index.html
```diff
@@ -1,28 +1,33 @@
...
-	</style>
-</head>
-<body class="d2l-typography">
+	<head>
+		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
+		<script type="module">
+			import '@brightspace-ui/core/components/colors/colors.js';
+			import '@brightspace-ui/core/components/typography/typography.js';
+		</script>
+		<title>core-ui</title>
```
svanherk, line 9:
```suggestion
		<title>labs</title>
```

demo/components/opt-in-flyout/index.html
```diff
@@ -0,0 +1,101 @@
...
+							help-docs-link="https://www.example.com#documentation">
+						</d2l-labs-opt-out-flyout>
+					</div>
+				</template>
+			</d2l-demo-snippet>
+
+			<script type="module">
+				[...document.querySelectorAll('d2l-labs-opt-in-flyout, d2l-labs-opt-out-flyout')].forEach(elem => {
+					['flyout-opened', 'flyout-closed', 'opt-in', 'opt-out'].forEach(eventName => {
+						// eslint-disable-next-line no-console
```
svanherk, line 94: Minor: Might be worth setting up an exception in `.eslintrc.json` for the demo folder like `core` does

src/components/localize-labs-element.js
svanherk, line 1: So if they all share a language file, I think we'll need to make a decision about OSLO - either _all_ terms are going to be OSLO-ified in a pretty weird organizational spot in the LMS, or we say we don't want Labs components to allow overrides. Only one Labs component does so currently.

Or I guess if you need OSLO, then you create your own lang file just for that component?

src/components/opt-in-flyout/README.md
```diff
@@ -0,0 +1,59 @@
+# Opt-in/Opt-out Flyouts
+
+The `<d2l-labs-opt-in-flyout>` and `<d2l-labs-opt-out-flyout>` can be used in applications to enable users to opt-in or out of new experiences.
+
+<!-- docs: start hidden content -->
```
svanherk, line 5: Is this repo hooked up to the Daylight site?  Are you looking to do that as part of this work, or just setting things up nicely in case we one day decide to?

src/components/localize-labs-element.js
dlockhart: It's a good point. In theory Labs should get the same exemption to OSLO as core... but even if core and labs needed to support OSLO one day, I don't think having them live in `web-components/core` and `web-components/labs` would be too strange? Or yeah, we could split labs up.

svanherk: It just seems weird to me for a client to have to know our flyout is a "labs" component (just like it's weird that they have to know it's a web component at all), and kinda gives away that we don't consider that piece of the LMS as polished lol. I guess I'd like to rework how the terms organize anyways, so maybe we make sure to do that _before_ exposing OSLO.  We'll need to see if anyone has overridden the `attachment` langterms before removing the ability too

dbatiste: Currently the lang-terms are written as `components:optInFlyout:*`.

Dave, you renamed the components to be `d2l-labs-*` in a previous commit. Is the above lang-term prefix still what you want? 

And if we do support overrides for labs components, what should we be using as a prefix in order to make migration to non-labs easier? On one hand if we use the above, maybe no migration is necessary, although that format is not quite what core uses. If we do have to change lang-terms, it seems like overrides will make migrating the labs component harder.

svanherk:
Side note - I think if we're going to have an influx of reviews here as components are moved in, setting up the demo page preview sites would be super awesome

src/components/localize-labs-element.js
dlockhart: If we cared about that yeah, although I don't think we've ever done that for anything else in the past -- including FRA migrations.

PR #100, opened 2023-08-14
rename title -> flyout-title and open -> opened

dlockhart:
While writing the vdiff tests, I noticed that because the attribute is named "title", the flyouts get a permanent browser tooltip whenever you hover over anywhere within the flyout. So this renamed that to `flyout-title` similar to how collapsible panel works.

Since we're dealing with breaking changes anyway, I also renamed `open` -> `opened` to match our dialogs.


dbatiste (approved):
Webkit tests timed out. Reran and all looks good!

PR #102, opened 2023-08-14
add vdiff tests for opt-in-flyout

dlockhart:


.github/workflows/update-package-lock.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - uses: Brightspace/third-party-actions@actions/setup-node
         with:
           node-version-file: .nvmrc
-          cache: npm
+          cache: 'npm'
```
dlockhart, line 18: The docs for this action have quotes around this, so making this consistent.

test/components/opt-in-flyout/opt-in-flyout.test.js
dlockhart, line 1: The vdiff tests make the vast majority of these redundant, so I've removed everything except the tests that assert the correct events are fired.

test/components/opt-in-flyout/opt-in-flyout.vdiff.js
```diff
@@ -0,0 +1,156 @@
+import '../../../src/components/opt-in-flyout/flyout-impl.js';
+import '../../../src/components/opt-in-flyout/opt-in-flyout.js';
+import '../../../src/components/opt-in-flyout/opt-out-flyout.js';
+import { expect, fixture, html, oneEvent, sendKeysElem } from '@brightspace-ui/testing';
+
+export function wrap(elem) {
+	return html`<div style="height: 500px; position: relative; width: 100%;">${elem}</div>`;
```
dlockhart, line 7: This wrapper was required in the same way it was needed for the demos because otherwise the flyout has no size. I couldn't just grab the whole body/viewport either because it also had no size.

bearfriend (approved):

test/components/opt-in-flyout/opt-in-flyout.vdiff.js
svanherk: Even using the new fullscreen support by passing in `document`?

dlockhart: 🤦 I tried it and it didn't work, but now trying again I was on version `0.28` of the testing library which didn't have your update. Fixed!

svanherk: Nice! The added margin at the top looks a bit weird in the closed case, but I'm guessing that's because the page margins affect the component, which is more of a bug in the component than in the test itself?

svanherk (approved):

test/components/opt-in-flyout/opt-in-flyout.vdiff.js
dlockhart: Yeah exactly. It's expecting to be placed inside something that's `position: relative`, which doesn't exist in these empty pages.

PR #111, opened 2023-08-21
fix: add a11y tests and fix select label

dlockhart:
If this is meant to act as an example, I figured I should add some accessibility tests. They found a missing label on a `<select>`. There's a bunch of other weird `<label>` stuff going on here, and they don't match our actual for label styles, but I think I'm going to draw the line here to limit visual changes.

### DIFF
--- a/src/components/opt-in-flyout/opt-out-reason-selector.js
+++ b/src/components/opt-in-flyout/opt-out-reason-selector.js
@@ -20,6 +20,11 @@ class OptOutReasonSelector extends composeMixins(
 		return [
 			inputStyles,
 			css`
+				label {
+					display: block;
+					margin-bottom: 0.5rem;
+				}
+
 				select {
 					-moz-appearance: none;
 					-webkit-appearance: none;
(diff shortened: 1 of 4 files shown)


src/components/opt-in-flyout/opt-out-reason-selector.js
```diff
@@ -68,6 +73,7 @@ class OptOutReasonSelector extends composeMixins(
 
 	render() {
 		return html`
+			<label for="selector">${this.localize('components:optInFlyout:feedbackReasonLabel')}</label>
```
dbatiste, line 76: Should this be getting the `d2l-input-label` class like the core input labels do?

dlockhart: So you can see what this actually looks like:

Screenshot 2023-08-21 at 2 50 45 PM

The answer is "probably"? I started going down that path, and then converting the `<d2l-input-textarea>` to also use our built-in labels, but they end up being `0.8rem` instead of `1rem` so visually things look quite different. As I mentioned in the PR description, I didn't really feel comfortable changing the visual style as part of what should just be a migration to the monorepo.

dbatiste: Yep, fair enough.

dbatiste (approved):

PR #129, opened 2023-09-07
Community Links

NicholasHallman:
## Implementation Note

There are a [few places where we use maps to direct users to a localized community link](https://github.com/Brightspace/insights-engagement-dashboard/blob/master/components/community-help.js). The goal of this component is to standardize this pattern and make it easy to use as an MVC control (coming soon) 

This component extends core links and adds the ability to dynamially navigate to a brightspace community article given an article code to lang code map.

### Functional
 - exact match goes to proper article
   - en
   - es-MX
   - pt-BR
   - nl-NL
   - ar-SA
   - fr-CA
 - different regions / no region codes match
   - en
   - es
   - pt
   - nl
   - ar
   - fr


src/components/community-link/community-link.js
```diff
@@ -0,0 +1,52 @@
+import '@brightspace-ui/core/components/link/link.js';
+
+import { getDocumentLocaleSettings } from '@brightspace-ui/intl/lib/common.js';
+
+const Link = customElements.get('d2l-link');
```
dlockhart, line 5: Hmm, interesting! I think I'd rather either:

1. We add an `export` to link's file for the class so you can just `import { LinkClass } from '@brightspace-ui/core/components/link/link.js';`. More realistically we should probably split it into two files -- one that defines the class and one that imports it and registers the custom element. That way you could import it without registering.
2. You just expose whatever of link's properties you want here, import `linkStyles` and apply them to your own `<a class="d2l-link">` element.

If you don't need all the crazy link properties, I'd probably lean towards 2. It's already a bit weird that the consumer of this component can set the `href` and then it'll get blown away.

It would also give you the opportunity to reign in what consumers can do if you like. Like do you want them to be able to put _anything_ inside the link? If not, you could add a `text` property and just always render that. Will the links _always_ open in a new window? Then don't even let them set `target` and just always apply that yourself. And so on.

src/components/community-link/community-link.js
```diff
@@ -0,0 +1,52 @@
...
+			if (articleCode === undefined) {
+				articleCode = this.articleMap['en'];
+				usedLangCode = 'en';
+			}
+		}
+		return { articleCode, langCode: usedLangCode };
+	}
+}
+
+customElements.define('d2l-community-link', CommunityLink);
```
dlockhart, line 52: All our labs components should be prefixed with `d2l-labs-*`.

src/components/community-link/community-link.js
```diff
@@ -0,0 +1,52 @@
...
+		const { language } = getDocumentLocaleSettings();
+		const { articleCode, langCode } = this._getArticleCode(language);
+
+		return `https://community.d2l.com/brightspace${langCode === 'en' ? '' : `-${langCode}`}/kb/articles/${articleCode}`;
+	}
+
+	render() {
+		this.href = this.link;
+		return super.render();
+	}
```
dlockhart, line 26: I'd suggest something like the following. Otherwise when the language on the page changes you won't be notified.

```suggestion
	constructor() {
		super();
		this._localeSettings = getDocumentLocaleSettings();
	}
	connectedCallback() {
		super.connectedCallback();
		this._localeSettings.addChangeListener(this._onLangChange);
	}
	disconnectedCallback() {
		super.disconnectedCallback();
		this._localeSettings.removeChangeListener(this._onLangChange);
	}
	render() {
		const { articleCode, langCode } = this._getArticleCode(this._localeSettings.language);
		const href = `https://community.d2l.com/brightspace${langCode === 'en' ? '' : `-${langCode}`}/kb/articles/${articleCode}`;
		return html`<a href="${href}" class="d2l-link"><slot></slot></a>`;
	}
	_onLangChange() {
		this.requestUpdate();
	}
```

src/components/community-link/community-link.js
dlockhart: While I'm nitpicking the name, typically we'll group similar elements together alphabetically so I'd suggest `<d2l-labs-link-community>`. (e.g. `d2l-input-date`, `d2l-input-time`, `d2l-button-subtle`, `d2l-button-icon`, etc.).

src/components/community-link/community-link.js
```diff
@@ -0,0 +1,52 @@
...
+import { getDocumentLocaleSettings } from '@brightspace-ui/intl/lib/common.js';
+
+const Link = customElements.get('d2l-link');
+
+class CommunityLink extends Link {
+
+	static get properties() {
+		return {
+			...Link.properties,
+			articleMap: { attribute: 'article-map', type: Object }
```
dlockhart, line 12: Yeah not sure how I feel about this API but can't really think of anything better. 🤔

demo/components/community-link/index.html
```diff
@@ -0,0 +1,37 @@
...
+	<head>
+		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
+		<script type="module">
+			import '@brightspace-ui/core/components/demo/demo-page.js';
+			import '../../../src/components/community-link/community-link.js';
+		</script>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta charset="UTF-8">
+		<style>
+			.flyout-wrapper {
```
dbatiste, line 12: This looks like it was copy & pasted but not needed.

src/components/community-link/community-link.js
NicholasHallman: Yeah unfortunatly the new community doesn't really make this easy...

NicholasHallman: alternatively it could use data-lang-<code> like
```
<d2l-labs-link-community
  data-lang-en="11064"
  data-lang-fr-CA="89026"
></d2l-labs-link-community>
```

src/components/community-link/community-link.js
NicholasHallman: Yeah my thinking here is that it would be a true "inheritor" of d2l-link. My first implementation was just taking and passing all the parent community link properties into a child link which seemed silly.

> It's already a bit weird that the consumer of this component can set the href and then it'll get blown away.

I could remove href as a property?

> Will the links always open in a new window?

For our use case yes, but I don't know about for all uses. Using the component to set a standard (all community links open in a new tab) maybe is desired?

>  Like do you want them to be able to put anything inside the link?

Could be a button? For insights dashboard community links they are a help button. Being able to slot in a button would be pretty handy

src/components/community-link/community-link.js
dlockhart: Yeah that feels more difficult for consumers which will likely have the map handy anyway eh. The only other thing I can think of is if we centralized all the community links somewhere behind an API. Then consumers could go `<d2l-labsl-link-community key="engagement-dashboard-help">`, the component would reach out to the API (passing the active language) and the API would return the URL. _Definitely_ more complicated.

src/components/community-link/community-link.js
dlockhart: Semantically it should always be a link and not a button, since links can be opened in new tabs, bookmarked, etc. whereas buttons cannot. _Visually_ it could still look like a button, although we still try really hard to avoid that.

For me this is making more and more of an argument for this not inheriting from `Link` and just let it define its own properties. `<d2l-labs-link-community type="button|link" text="Get Help on Something">`.

NicholasHallman: That a good point... Okay I'm convinced, I'll make those change

src/components/community-link/community-link.js
NicholasHallman: Yeah I don't think even community has this api. It doesn't seem to understand translated articles are "the same" given that switching the translation on an article will just redirect you to the homepage in the new language

index.html
```diff
@@ -22,6 +22,7 @@ <h1 class="d2l-heading-2">Demos</h1>
 		<h2 class="d2l-heading-2">Components</h2>
 		<ul>
 			<li><a href="demo/components/opt-in-flyout/index.html">Opt-In Flyout</a></li>
+			<li><a href="demo/components/community-link/index.html">Community Link</a></li>
```
dbatiste, line 25: Nit: I think we'll want to put these in alpha order like we do for core.

src/components/community-link/community-link.js
```diff
@@ -0,0 +1,81 @@
...
+	static get properties() {
+		return {
+			text: { type: String },
+			type: { type: String },
+			articleMap: { attribute: 'article-map', type: Object }
+		};
+	}
+
+	constructor() {
+		super();
```
dlockhart, line 17:
```suggestion
		super();
		this.type = 'link';
```

src/components/community-link/community-link.js
```diff
@@ -0,0 +1,81 @@
...
+		this._localeSettings.removeChangeListener(this._handleUpdate);
+	}
+
+	render() {
+		return this.type === 'button'
+			? html`<d2l-button-subtle
+				icon="d2l-tier1:help"
+				text=${this.text}
+				@click="${this._handleButtonClick}"
+			>`
```
dlockhart, line 45: Custom elements always need a closing tag.
```suggestion
			></d2l-button-subtle>`
```

src/components/community-link/community-link.js
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this._localeSettings.removeChangeListener(this._handleUpdate);
+	}
+
+	render() {
+		return this.type === 'button'
+			? html`<d2l-button-subtle
```
dlockhart, line 41: So unfortunately by using subtle buttons (which render native `<button>`), the link semantics are gone. The ideal scenario here would be a `<a>` that's styled to look like a button, but like I said we still try to avoid this whenever possible since the user loses all the link niceties like right-clicking.

But this _is_ a labs component so I don't want to get super strict in here...

src/components/community-link/community-link.js
```diff
@@ -0,0 +1,81 @@
+import '@brightspace-ui/core/components/link/link.js';
+import '@brightspace-ui/core/components/button/button-icon.js';
```
dlockhart, line 2: This is the wrong import.
```suggestion
import '@brightspace-ui/core/components/button/button-subtle.js';
```

src/components/community-link/community-link.js
```diff
@@ -0,0 +1,81 @@
...
+			}
+		}
+		return { articleCode, langCode: usedLangCode };
+	}
+
+	_handleButtonClick() {
+		window.open(this.link, '_blank');
+	}
+
+	_onLangChange() {
```
dlockhart, line 76: Don't think this ended up getting used?

src/components/community-link/community-link.js
dlockhart, line 1: Let's rename the directories and filenames to match the component -- `link-community`.

Once we're happy with this, you'll want to add it to the exports in `package.json`, add a README and some vdiff tests. This repo is configured to use the new vdiff framework (lucky you!). You should be able to use the flyout ones as a guide, but if you need more help reach out to myself or [the docs are here](https://github.com/BrightspaceUI/testing).

src/components/community-link/community-link.js
NicholasHallman: Yeah it's not opimal. I copied this pattern from the insights dashboards where it looks like this

community link

NicholasHallman: It's a subtle button because the overflow panel shows the button witht he icon (sometimes lol)

NicholasHallman:
Option #3

Create a community article lang map directive that produces the links and give the `d2l-labs-link-community` and `d2l-labs-button-community` as prebuilt examples so I can use it in the lms easier?

NicholasHallman:
Okay I've gone ahead and split the logic into a few pieces. The idea being that, regardless of how you want to get to community (button, link, etc) it should be easy to do so with it's translated siblings. So the `communityUrlFactory` will generate a directive for you (given an article map) and the resulting directive will generate the URL (given a language) for use however you want. The two example components (link, and button) exist so that I can go ahead and add this as an MVC controller cause I'm not sure if I can use the directive as is (maybe a future wasm / ssr project?)

src/controllers/language-listener/language-listener.js
```diff
@@ -0,0 +1,23 @@
...
+	hostConnected() {
+		this._localeSettings.addChangeListener(this._handleLangUpdate);
+	}
+
+	hostDisconnected() {
+		this._localeSettings.removeChangeListener(this._handleLangUpdate);
+	}
+
+	hostUpdate() {
+		this.language = this._localeSettings.language;
```
dlockhart, line 21: Couldn't this be set inside `_handleLangUpdate` and then there'd be no need for `hostUpdate`? I've only seen `hostUpdate` used to poke at the host's DOM before it's updated.

src/controllers/language-listener/language-listener.js
dlockhart, line 1: Cool, I really like this! Solves the problem of needing to know when the language changes when `LocalizeMixin` isn't in use.

src/components/community/link-community.js
```diff
@@ -0,0 +1,31 @@
+import '@brightspace-ui/core/components/link/link.js';
+import '@brightspace-ui/core/components/button/button-subtle.js';
```
dlockhart, line 2: No need to import subtle button anymore.

src/components/community/button-community.js
```diff
@@ -0,0 +1,40 @@
+import '@brightspace-ui/core/components/link/link.js';
```
dlockhart, line 1: No need to import link here.

src/components/community/button-community.js
```diff
@@ -0,0 +1,40 @@
...
+import { LanguageListenerController } from '../../controllers/language-listener/language-listener.js';
+
+class CommunityLink extends LitElement {
+
+	static get properties() {
+		return {
+			text: { type: String },
+			articleMap: { attribute: 'article-map', type: Object }
+		};
+	}
```
dlockhart, line 14: If you want, you could create a base component that has the properties, constructor with `langController` setup and `connectedCallback` and then both would extend that and just need to provide `render()`.

src/components/community/directive/community-article.js
dlockhart, line 1: I don't think this need to be inside a `directive` directory? Could just be `url-factory.js`.

src/components/community/link-community.js
```diff
@@ -0,0 +1,31 @@
...
+		super.connectedCallback();
+		this.communityArticleDirective = communityUrlFactory(this.articleMap);
+	}
+
+	render() {
+		return html`<d2l-link href="${this.communityArticleDirective(this.langController.language)}" target="_blank">${this.text}</d2l-link>`;
+	}
+}
+
+customElements.define('d2l-labs-link-community', CommunityLink);
```
dlockhart, line 31: You're going to hate me, but now that these are both "community" things in a "community" directory, it's making me want to flip-flop and have them to be `d2l-labs-communit-link` and `d2l-labs-community-button`.

The files would then become `community-button.js` and `community-link.js`.

test/components/link-community/link-community.test.js
```diff
@@ -0,0 +1,53 @@
...
+	'pt-BR': 'https://community.d2l.com/brightspace-pt-BR/kb/articles/9978',
+	'nl-NL': 'https://community.d2l.com/brightspace-nl-NL/kb/articles/7833',
+	'fr-CA': 'https://community.d2l.com/brightspace-fr-CA/kb/articles/6739'
+};
+
+describe('community link', () => {
+	Object.keys(langArticleMap).map((langCode) => {
+		it(`Should have the right link for each lang code ${langCode}`, async() => {
+			getDocumentLocaleSettings().language = langCode;
+			const communityLink = await fixture(html`<d2l-labs-link-community type="link" text="Help" article-map=${JSON.stringify(langArticleMap)}></d2l-labs-link-community>`);
```
dlockhart, line 28: Just pass the language into `fixture`:

```suggestion
			const communityLink = await fixture(html`<d2l-labs-link-community type="link" text="Help" article-map=${JSON.stringify(langArticleMap)}></d2l-labs-link-community>`, { lang: langCode });
```

test/components/link-community/link-community.test.js
dlockhart, line 1: I'd suggest all these unit tests just test the directive directly -- no need to tie them to the web component now that the logic has moved out of it.

test/components/link-community/link-community.test.js
```diff
@@ -0,0 +1,53 @@
...
+		});
+	});
+
+	it("Should default to english if the lang code doesn't match", async() => {
+		getDocumentLocaleSettings().language = 'zh-tw';
+		const communityLink = await fixture(html`<d2l-labs-link-community type="link" text="Help" article-map=${JSON.stringify(langArticleMap)}></d2l-labs-link-community>`);
+		expect(communityLink.shadowRoot.querySelector('d2l-link').href).to.equal(langUrlMap['en']);
+	});
+
+	it('Should have a button variant', async() => {
```
dlockhart, line 48: I don't think this test is necessary -- the vdiff handles this.

test/components/link-community/link-community.vdiff.js
```diff
@@ -0,0 +1,37 @@
...
+const opts = { viewport: { height: 450 } };
+
+const langArticleMap = {
+	'en': 4545,
+	'ar-SA': 10489,
+	'es-MX': 9214,
+	'pt-BR': 9978,
+	'nl-NL': 7833,
+	'fr-CA': 6739
+};
```
dlockhart, line 15: Can you `export` this from somewhere common and share it between all your tests and the demo?

test/components/link-community/link-community.vdiff.js
```diff
@@ -0,0 +1,37 @@
...
+
+describe('link-community', () => {
```
dlockhart, line 17:
```suggestion
describe('community', () => {
```

test/components/link-community/link-community.vdiff.js
```diff
@@ -0,0 +1,37 @@
...
+	it('Should render a link in link mode', async() => {
```
dlockhart, line 18: Since these names end up in the filename of the diff file, I'd suggest keeping it simple:

```suggestion
	it('link', async() => {
```

test/components/link-community/link-community.vdiff.js
```diff
@@ -0,0 +1,37 @@
...
+		await fixture(
+			html`<d2l-labs-link-community
+				article-map="${JSON.stringify(langArticleMap)}"
+				text="Link To Community"
+			></d2l-labs-link-community>`
+			, opts);
+		await expect(document).to.be.golden();
```
dlockhart, line 25: Any particular reason you're snapshotting the whole document instead of just the component?

src/components/community/link-community.js
NicholasHallman: Sounds good and I get it haha it makes sense

test/components/link-community/link-community.test.js
NicholasHallman: Neat!

demo/controllers/language-listener/index.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
+		<script type="module">
+			import '@brightspace-ui/core/components/colors/colors.js';
```
dlockhart, line 6: Don't think you need this.
```suggestion
```

src/components/community/community-button.js
```diff
@@ -0,0 +1,28 @@
+import '@brightspace-ui/core/components/button/button-subtle.js';
+import { CommunityBase } from './community-base.js';
+import { html } from 'lit';
+
+class CommunityLink extends CommunityBase {
+
+	static get properties() {
+		return {
+			...CommunityBase.properties,
```
dlockhart, line 9: I'd suggest doing `CommunityBase` as a Lit mixin instead -- that way these inherited classes don't need to add its properties to theirs.

src/components/community/community-base.js
```diff
@@ -0,0 +1,22 @@
...
+	}
+
+	constructor() {
+		super();
+		this.langController = new LanguageListenerController(this);
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		this.communityArticleDirective = communityUrlFactory(this.articleMap);
```
dlockhart, line 20: I think this is going to result in the map getting resolved on first connection and if the consumer changes `articleMap` (via property or attribute) it will never get updated. To fix, you could either do this in `updated` whenever `articleMap` changes?

src/controllers/language-listener/language-listener.js
dlockhart: Probably worth adding a few simple unit tests to cover the main functionality.

dlockhart (commented):
Looking really good! Just a few additional comments.

test/components/link-community/link-community.vdiff.js
bearfriend: Add this to the list of language maps 😄

src/controllers/language-listener/language-listener.js
```diff
@@ -0,0 +1,22 @@
...
+	}
+
+	hostConnected() {
+		this._localeSettings.addChangeListener(this._handleLangUpdate);
+	}
+
+	hostDisconnected() {
+		this._localeSettings.removeChangeListener(this._handleLangUpdate);
+	}
+}
```
bearfriend, line 22: Not a big deal but seems clearer to me that each instance doesn't actually have its own locale settings this way:
```suggestion
const localeSettings = getDocumentLocaleSettings();

export class LanguageListenerController {

	constructor(host) {
		(this.host = host).addController(this);
		this._handleLangUpdate = () => {
			this.language = localeSettings.language;
			this.host.requestUpdate();
		};
		this.language = localeSettings.language;
	}

	hostConnected() {
		localeSettings.addChangeListener(this._handleLangUpdate);
	}

	hostDisconnected() {
		localeSettings.removeChangeListener(this._handleLangUpdate);
	}
}
```

bearfriend: Yikes, that's not a helpful diff at all. Just broke `this._localeSettings` outside the class to `const localeSettings = getDocumentLocaleSettings();` and referenced that within the class.

src/components/community/community-base.js
```diff
@@ -19,4 +18,9 @@ export class CommunityBase extends LitElement {
 		super.connectedCallback();
 		this.communityArticleDirective = communityUrlFactory(this.articleMap);
 	}
-}
+
+	update() {
+		super.update();
+		this.communityArticleDirective = communityUrlFactory(this.articleMap);
+	}
```
dlockhart, line 25: Should be `updated` (not `update`) and this should get called always for new values as well so you don't need to set it in the constructor.

```suggestion 
	updated(changedProperties) {
		if (changedProperties.has('articleMap')) {
			this.communityArticleDirective = communityUrlFactory(this.articleMap);
		}
	}
```

test/controllers/language-listener/language-listener.test.js
```diff
@@ -0,0 +1,23 @@
...
+	it('Should get the current language', () => {
+		const host = { addController: () => {} };
+		const langListener = new LanguageListenerController(host);
+		expect(langListener.language).to.equal('en');
+	});
+
+	it('Should change based on the document language', async() => {
+		// using en example element here for host updating features
+		const elm = await fixture(html`<d2l-labs-community-link article-map="${JSON.stringify(langArticleMap)}"></d2l-labs-community-link>`);
+		localeSettings.language = 'fr';
```
dlockhart, line 20: The new testing library allows for passing in the languages:

```suggestion
		const elm = await fixture(html`<d2l-labs-community-link article-map="${JSON.stringify(langArticleMap)}"></d2l-labs-community-link>`, { lang: 'fr' });
```

test/controllers/language-listener/language-listener.test.js
```diff
@@ -0,0 +1,23 @@
+import '../../../src/components/community/community-link.js';
```
dlockhart, line 1: I'd suggest using the lang-listener demo component or just making one on-the-fly using `defineCE`. That way the unit tests aren't relying on another component which may change, move or be removed in the future.

test/controllers/language-listener/language-listener.test.js
NicholasHallman: Oh yes I can add the language to the fixture but this test is making sure that it updates after first render

dlockhart: Cool yep my bad!

test/controllers/language-listener/language-listener.test.js
```diff
@@ -2,22 +2,29 @@ import '../../../src/components/community/community-link.js';
...
+import { LitElement } from 'lit';
 
 const localeSettings = getDocumentLocaleSettings();
 
+customElements.define('example-lang-listener', class extends LitElement {
+	constructor() {
+		super();
+		this.langController = new LanguageListenerController(this);
+	}
+});
```
dlockhart, line 15: Alternative using `defineCE`:

```javascript
const tag = defineCE(
	class extends LitElement {
		constructor() {
			super();
			this.langController = new LanguageListenerController(this);
		}
	}
);
```

And then:

```javascript
const elem = await fixture(`<tag></tag>`, { lang: 'fr' });
```

NicholasHallman:
I got a little hasty and merged in the link with the `new-window` change from [here](https://github.com/BrightspaceUI/core/pull/4001). I'll open a new pr if new-window doesn't end up being the attribute name

PR #198, opened 2023-12-18
US159909 Move checkbox-drawer to labs monorepo

ChrisLabattD2L:
Context for [US159909](https://rally1.rallydev.com/#/?detail=/userstory/730793441071&fdp=true): Migrate Checkbox-Drawer to the Labs Monorepo

The purpose of this PR is to move the code from [`checkbox-drawer` repo](https://github.com/BrightspaceUILabs/checkbox-drawer) to the labs monorepo with the goal of eventually graveyarding the old repo once it's been move over and all of the references to the old repo are update to point to here instead.

There are a few changes between the code in the old repo and here, which include:
- Simplifying and reusing code where possible
- Replacing some of the old tests with visual-diff tests, since the old repos didn't initially have them.
- Updating lang term names to match the format used in this repo

I've added the component to the package.json exports section, not sure if there's anything else that might need to be done to initialize the component in this repo since I couldn't find documentation regarding what needs to be done in this repo.


src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -0,0 +1,112 @@
...
+import { bodyCompactStyles } from '@brightspace-ui/core/components/typography/styles.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { LocalizeLabsElement } from '../localize-labs-element.js';
+import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
+
+class CheckboxDrawer extends LocalizeLabsElement(SkeletonMixin(LitElement)) {
+
+	static get properties() {
+		return {
+			ariaLabel: { type: String, attribute: 'aria-label' },
```
dlockhart, line 16: There seem to only be a few usages of this component -- would you be open to renaming this attribute while switching things over? We're not supposed to use the `aria-*` attributes on custom elements, which was a mistake we made in `<d2l-input-checkbox>` but is much harder to correct at this point.

Actually... combing through the search results, I don't think `aria-label` is used at all with this, so maybe it can just go away?

src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -0,0 +1,112 @@
...
+				</d2l-expand-collapse-content>
+			</d2l-input-checkbox-spacer>
+		`;
+	}
+
+	_onCheckboxChange(e) {
+		this.checked = e.target.checked;
+
+		this.dispatchEvent(
+			new CustomEvent('d2l-checkbox-drawer-checked-change', {
```
dlockhart, line 86: Ideally these event names have `-labs-` in them -- if you're open to making that breaking change.

src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -0,0 +1,112 @@
...
+	async _onExpandCollapseContentToggled(e) {
+		const action = e.target.expanded ? 'expand' : 'collapse';
+
+		this.dispatchEvent(
+			new CustomEvent(`d2l-checkbox-drawer-${action}`, {
+				bubbles: true,
+				composed: false,
+				detail: e.detail
+			})
+		);
```
dlockhart, line 103: Nit: could you do something like this to make it easier to code search for the event names?

```suggestion
		const eventName = e.target.expanded ? `d2l-checkbox-drawer-expand` : `d2l-checkbox-drawer-collapse`;
		this.dispatchEvent(
			new CustomEvent(eventName, {
				bubbles: true,
				composed: false,
				detail: e.detail
			})
		);
```

src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -0,0 +1,112 @@
...
+
+		const content = this.shadowRoot?.querySelector('d2l-expand-collapse-content');
+		content.ariaBusy = 'true';
```
dlockhart, line 106: Ideally setting this attribute happens in `render()` otherwise you're modifying things outside of the render loop -- you could use a `state: true` property.

test/components/checkbox-drawer/checkbox-drawer.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+	describe('events', () => {
+		[ false, true ].forEach(checked => {
+			const action = checked ? 'collapse' : 'expand';
+
+			it(`should fire "d2l-checkbox-drawer-checked-change" event when ${!checked ? 'un' : ''}checked input element is clicked`, async() => {
+				const component = await fixture(html`<d2l-labs-checkbox-drawer label="Label" ?checked=${checked}></d2l-labs-checkbox-drawer>`);
+				const listener = oneEvent(component, 'd2l-checkbox-drawer-checked-change');
+
+				getInputCheckbox(component).simulateClick();
```
dlockhart, line 34: You can use the native mouse events now:

```suggestion
				clickElem(getInputCheckbox(component));
```

test/components/checkbox-drawer/checkbox-drawer.test.js
```diff
@@ -0,0 +1,53 @@
...
+				const { detail } = await listener;
+				expect(detail.checked).to.equal(!checked);
+				expect(component.checked).to.equal(!checked);
+
+				const innerCheckbox = component.shadowRoot.querySelector('d2l-input-checkbox');
+				expect(innerCheckbox.description).to.equal(checked ? 'Checkbox collapsed' : 'Checkbox expanded');
```
dlockhart, line 40: This assertion feels like something that would be better suited to a vdiff.

test/components/checkbox-drawer/checkbox-drawer.test.js
```diff
@@ -0,0 +1,53 @@
...
+			});
+
+			it(`should fire "d2l-checkbox-drawer-${action}" event when ${!checked ? 'un' : ''}checked input element is clicked`, async() => {
+				const component = await fixture(html`<d2l-labs-checkbox-drawer label="Label" ?checked=${checked}></d2l-labs-checkbox-drawer>`);
+				const listener = oneEvent(component, `d2l-checkbox-drawer-${action}`);
+
+				getInputCheckbox(component).simulateClick();
```
dlockhart, line 47:
```suggestion
				clickElem(getInputCheckbox(component));
```

dlockhart (commented):
Thanks for converting this repo over! One fewer repos to manage is always a plus.

src/components/checkbox-drawer/checkbox-drawer.js
ChrisLabattD2L: Since this component is only ever used in my team's code in foundation-components ([here](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/main/features/discover/d2l-discover-rules.js#L202) and [here](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/main/features/discover/d2l-discover-approval-flows-config.js#L100)), I don't mind making this change along with any other breaking changes (like the one you mentioned in your other comment below). So if there are any other such changes that would help bring the component more in line with other labs or core component, just let me know and I can make those changes too 👍

src/components/checkbox-drawer/README.md
```diff
@@ -0,0 +1,50 @@
...
+|--|--|
+| `aria-label` | Provides context for the component. Must be used if `label` is not used. Cannot be used with `label`. |
+
+**Events:**
+
+- `d2l-checkbox-drawer-checked-change`: dispatched when checkbox's state changes.
+- `d2l-checkbox-drawer-expand`: dispatched when the drawer starts to expand. As per the [expand collapse component](https://github.com/BrightspaceUI/core/tree/master/components/expand-collapse), the `detail` contains an `expandComplete` promise that can be waited on to determine when the content has finished expanding.
+- `d2l-checkbox-drawer-collapse`: dispatched when the drawer starts to collapse. As per the [expand collapse component](https://github.com/BrightspaceUI/core/tree/master/components/expand-collapse), the `detail` contains a `collapseComplete` promise that can be waited on to determine when the content has finished collapsing.
+
+> Note: this is a ["labs" component](https://daylight.d2l.dev/developing/getting-started/component-tiers/). While functional, these tasks are prerequisites to promotion to BrightspaceUI "official" status:
```
dlockhart, line 39: You can remove this section from the README now -- I don't think we need to track this with each component anymore.

src/components/checkbox-drawer/checkbox-drawer.js
dlockhart: Amazing, thanks! Yeah I think I'd just remove this attribute and then update the event names to match the component name.

test/components/checkbox-drawer/checkbox-drawer.test.js
ChrisLabattD2L: In this case `description` is an offscreen value that's used for `aria-describedby` attribute in the `d2l-input-checkbox` component and not the description in the checkbox-drawer. Is visual-diff able to test for offscreen components?

dlockhart: Ah ok my bad -- doing it here is fine.

src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -71,7 +70,7 @@ class CheckboxDrawer extends LocalizeLabsElement(SkeletonMixin(LitElement)) {
 				<div class="d2l-input-checkbox-description d2l-skeletize">${this.description}</div>
 			</d2l-input-checkbox-spacer>
 			<d2l-input-checkbox-spacer class="d2l-input-checkbox-spacer">
-				<d2l-expand-collapse-content @d2l-expand-collapse-content-expand="${this._onExpandCollapseContentToggled}" @d2l-expand-collapse-content-collapse="${this._onExpandCollapseContentToggled}" ?expanded="${this.checked}">
+				<d2l-expand-collapse-content ?expanded="${this.checked}" ?aria-busy=${this._expandCollapseBusy} @d2l-expand-collapse-content-expand="${this._onExpandCollapseContentToggled}" @d2l-expand-collapse-content-collapse="${this._onExpandCollapseContentToggled}">
```
dlockhart, line 73: This looks good, except `aria-busy` is oddly a `String`, so this should be:
```suggestion
				<d2l-expand-collapse-content ?expanded="${this.checked}" aria-busy=${this._expandCollapseBusy ? 'true' : 'false'} @d2l-expand-collapse-content-expand="${this._onExpandCollapseContentToggled}" @d2l-expand-collapse-content-collapse="${this._onExpandCollapseContentToggled}">
```

src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -92,20 +91,22 @@ class CheckboxDrawer extends LocalizeLabsElement(SkeletonMixin(LitElement)) {
...
 				detail: e.detail
 			})
 		);
 
-		const content = this.shadowRoot?.querySelector('d2l-expand-collapse-content');
-		content.ariaBusy = 'true';
-		await e.detail[`${action}Complete`];
-		content.ariaBusy = 'false';
+		this._expandCollapseBusy = true;
+		await this.updateComplete;
```
dlockhart, line 106: You shouldn't need this `await this.updateComplete` since you don't really care about waiting for the next render before awaiting the detail.

src/components/checkbox-drawer/checkbox-drawer.js
ChrisLabattD2L: Oh boy that's not confusing at all 😕 thanks for the heads up! I've changed that (I'll have to make sure the other places I've seen it in Discover are correct 😅 )

PR #201, opened 2024-01-05
US159929 Move Pagination to labs monorepo

ChrisLabattD2L:
Context for [US159929](https://rally1.rallydev.com/#/?detail=/userstory/730838419855&fdp=true): Migrate Pagination to the Labs Monorepo

The purpose of this PR is to move the code from [pagination labs repo](https://github.com/BrightspaceUILabs/pagination) to the labs monorepo with the goal of eventually graveyarding the old repo once it's been move over and all of the references to the old repo are update to point to here instead.

There are a few changes between the code in the old repo and here, which include:

- Simplifying and reusing code where possible
- Replaced the use of [`d2l-input-text`](https://github.com/BrightspaceUILabs/pagination/blob/main/pagination.js#L77) with [`d2l-input-number`](https://daylight.d2l.dev/components/numeric-inputs/#d2l-input-number), since the pagination component was initially created prior to the `d2l-input-number` existing.
- Changed the names of the events to match the `d2l-labs-*` format
- Improved the format of the `README` file
- Replacing some of the old tests with visual-diff tests, since the old repos didn't initially have them.
- Updating lang term names to match the format used in this repo
- Fixed an issue where the fraction was being flipped in RTL when it shouldn't be flipped (see [here](https://en.wikipedia.org/wiki/Modern_Arabic_mathematical_notation#:~:text=In%2Dline%20fractions%20are%20written%20with%20the%20numerator%20and%20denominator%20on%20the%20left%20and%20right%20of%20the%20fraction%20slash%20respectively%2C%20e.g.%20%D9%A2/%D9%A7%202/7.))

## Larger changes:

- Changed the component's name from `d2l-labs-pagination` to `d2l-labs-pager-numeric`
- Changed property names:
  - `showItemCountSelect` -> `showPageSizeSelector`
  - `itemCountOptions` -> `pageSizes`
  - `selectedCountOption` -> `pageSize`


demo/components/pagination/index.html
```diff
@@ -0,0 +1,44 @@
...
+
+			<h2>Item Count Enabled</h2>
+			<d2l-demo-snippet>
+				<d2l-labs-pagination max-page-number="5" show-item-count-select item-count-options="[1,10,20,100,250]" selected-count-option="10"></d2l-labs-pagination>
+			</d2l-demo-snippet>
+		</d2l-demo-page>
+
+		<script>
+			const elements = document.querySelectorAll('d2l-labs-pagination');
+			elements.forEach(elem => {
```
tenbradley, line 33: Thanks for condensing this.

src/components/pagination/README.md
```diff
@@ -0,0 +1,29 @@
...
+    import '@brightspace-ui-labs/pagination/pagination.js';
+</script>
+<d2l-labs-pagination></d2l-labs-pagination>
+```
+
+## Properties
+
+| Property | Type | Description |
+|--|--|--|
+| `page-number` | Number, default: 1 | The current page number
```
tenbradley, line 18: In the old repo, a couple of these were listed as required. Is that not stated in this repo, replaced by the default value, no longer true, or not here for some other reason?

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,143 @@
...
+			maxPageNumber: { type: Number, attribute: 'max-page-number', reflect: true },
+			showItemCountSelect : { type: Boolean, attribute: 'show-item-count-select', reflect: true },
+			itemCountOptions : { type: Array, attribute: 'item-count-options' },
+			selectedCountOption : { type: Number, attribute: 'selected-count-option', reflect: true },
+		};
+	}
+
+	static get styles() {
+		return [selectStyles, css`
+			:host {
```
tenbradley, line 23: A couple things were removed here. `align-content` and `flex-direction`. Are these not needed here?

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,143 @@
...
+				align-items: center;
+				display: flex;
+				justify-content: center;
+				white-space: nowrap;
+			}
+
+			.d2l-page-selector-container {
```
tenbradley, line 30: nit: alphabetical order might be mice here.

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,143 @@
...
+					@change=${this._onPageNumberChanged}>
+				</d2l-input-number>
+				<!-- Note: this uses a division slash rather than a regular slash -->
+				<!-- a11y note: setting aria-hidden to true because info here is covered by input element -->
+				<span class="d2l-page-max-text" aria-hidden="true">∕ ${this.maxPageNumber}</span>
+				<d2l-button-icon id="d2l-labs-pagination-next-button" icon="tier1:chevron-right" @click="${this._onNextClicked}" text="${this.localize('components:pagination:nextPage')}" ?disabled=${this._isLastPage()}></d2l-button-icon>
+			</div>
+
+			${this.showItemCountSelect ? html`
+				<select
```
tenbradley, line 80: Is there any benefit to select or the options having an `id`?

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,143 @@
...
+	_onPageNumberChanged(e) {
+		if (!this._isValidNumber(e.target.value)) {
+			e.target.value = this.pageNumber;
+			return;
+		}
+
+		this._dispatchPageChangeEvent(Number(e.target.value));
+	}
+
+	_onPreviousClicked() {
```
tenbradley, line 138: nit: unless this is so they are alphabetized, it might be nice to have this function right after/before the `_onNextClicked` function since they're related.

test/components/pagination/golden/pagination/chromium/arrow-button-states-rtl-previous-button-disabled-if-first-page.png
tenbradley, line 1: I'm not sure how this would apply when working with page numbers like this, but the internet, aka [wikipedia](https://en.wikipedia.org/wiki/Modern_Arabic_mathematical_notation), states that "In-line fractions are written with the numerator and denominator on the left and right of the fraction slash respectively, e.g. ٢/٧ 2/7" in Arabic. I don't know if this applies to all rtl languages, or if it applies in a case something like this, but figured I'd point it out just in case.

src/components/pagination/README.md
ChrisLabattD2L: So what I've noticed is that in other component cases, we don't actually have the `required` part in favour of saying what the default is since in theory the component still works without giving the `page-number`, but if that's done, you just won't be able to do much with it. I can add back the `required` field though, for both this and for the max page number

src/components/pagination/pagination.js
ChrisLabattD2L: In these cases, they were just assigning the default value, and therefore could get removed since they weren't providing any benefit

src/components/pagination/pagination.js
ChrisLabattD2L: From what I've noticed, the trend for the order for CSS classes tends to follow the flow of the code that's being written, which is why the order goes `d2l-page-selector-container` -> `d2l-page-number` -> `d2l-page-max-text` (with host being before it all and size-specific classes following afterwards)

src/components/pagination/pagination.js
ChrisLabattD2L: Yeah, because of linting, they have to appear in alphabetical order, otherwise it would definitely make sense to have them appear in that order

test/components/pagination/golden/pagination/chromium/arrow-button-states-rtl-previous-button-disabled-if-first-page.png
ChrisLabattD2L: Hmm, that's actually a pretty good point. I was mostly copying over what was already there, but if this is an improvement that can be potentially made, it's worth looking into. I'll ask about it in Slack to see what others say.

src/components/pagination/pagination.js
ChrisLabattD2L: Not really, since there's only a single select, it probably isn't worth having an id for it. On the other side, since the options are passed in by the user, I don't want to risk the array having duplicates and causing things to fail due to there being two components with the same ID (since they're meant to be unique)

test/components/pagination/golden/pagination/chromium/arrow-button-states-rtl-previous-button-disabled-if-first-page.png
ChrisLabattD2L: Here's a [link](https://d2l.slack.com/archives/C0PHG3QB0/p1704741860447299) to the post I made about this.

ChrisLabattD2L: Looks like the outcome is that the buttons should still change, but the fraction as you mentioned, should stay the same regardless. I'll make that change in this case 👍

tenbradley (approved):
Thanks Chris.

ChrisLabattD2L:
This PR should be ready for a @BrightspaceUI/gaudi-dev review!

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,150 @@
...
+import { selectStyles } from '@brightspace-ui/core/components/inputs/input-select-styles.js';
+
+class Pagination extends RtlMixin(LocalizeLabsElement(LitElement)) {
+
+	static get properties() {
+		return {
+			pageNumber: { type: Number, attribute: 'page-number', reflect: true },
+			maxPageNumber: { type: Number, attribute: 'max-page-number', reflect: true },
+			showItemCountSelect : { type: Boolean, attribute: 'show-item-count-select', reflect: true },
+			itemCountOptions : { type: Array, attribute: 'item-count-options' },
```
bearfriend, line 16: Does using the `item-count-options` attribute actually work? I think we'd need to add a custom converter, no? Maybe just not worth having the attribute at all.

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,150 @@
...
+						autocomplete="off"
+						label="${this.localize('components:pagination:currentPage', { pageNumber: this.pageNumber, maxPageNumber: this.maxPageNumber })}"
+						label-hidden
+						max-fraction-digits="0"
+						value="${this.pageNumber}"
+						@change=${this._onPageNumberChanged}>
+					</d2l-input-number>
+					<!-- Note: this uses a division slash rather than a regular slash -->
+					<!-- a11y note: setting aria-hidden to true because info here is covered by input element -->
+					<span class="d2l-page-max-text" aria-hidden="true">∕ ${this.maxPageNumber}</span>
```
dbatiste, line 81: I think we'd want to use `formatNumber` for this value. Here's the import you'll need:
```javascript
import { formatNumber } from '@brightspace-ui/intl/lib/number.js';
```

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,150 @@
...
+		super();
+		this.pageNumber = 1;
+		this.maxPageNumber = 1;
+		this.itemCountOptions = [10, 20, 30, 40];
+	}
+
+	render() {
+		return html`
+			<div class="d2l-page-selector-container">
+				<d2l-button-icon id="d2l-labs-pagination-previous-button" icon="tier1:chevron-left" @click="${this._onPreviousClicked}" text="${this.localize('components:pagination:previousPage')}" ?disabled=${this._isFirstPage()}></d2l-button-icon>
```
dbatiste, line 67: I think this could be `?disabled="${this.pageNumber <= 1}"`. Then we don't need that method. If the method were used elsewhere then it would be more useful. Not a big deal though. Same with the other button.

Note: I think Lit tolerates not having the double quotes, but our practice is to always use them.

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,150 @@
...
+
+				<div class="d2l-page-number-container">
+					<d2l-input-number
+						class="d2l-page-number"
+						autocomplete="off"
+						label="${this.localize('components:pagination:currentPage', { pageNumber: this.pageNumber, maxPageNumber: this.maxPageNumber })}"
+						label-hidden
+						max-fraction-digits="0"
+						value="${this.pageNumber}"
+						@change=${this._onPageNumberChanged}>
```
dbatiste, line 77: Nit: Missing quotes here.

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,150 @@
...
+			</div>
+
+			${this.showItemCountSelect ? html`
+				<select
+					class="d2l-input-select"
+					aria-label="${this.localize('components:pagination:resultsPerPage')}"
+					title="${this.localize('components:pagination:resultsPerPage')}"
+					@change="${this._onPageCounterChanged}">
+					${this.itemCountOptions.map(item => html`
+						<option ?selected="${this.selectedCountOption === item}" value="${item}">${this.localize('components:pagination:amountPerPage', 'count', item)}</option>
```
dbatiste, line 93: This might be ok if the numbers are always small, but I would recommend using `formatNumber` for the text that is displayed.

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,150 @@
...
+				</select>
+			` : null }
+		`;
+	}
+
+	_dispatchPageChangeEvent(newPageNumber) {
+		this.dispatchEvent(
+			new CustomEvent('d2l-labs-pagination-page-change', {
+				detail: { page: newPageNumber },
+				bubbles: true,
```
dbatiste, line 104: Again, not sure how disruptive this would be, but normally we would set `bubbles: false` and `composed: false`. We generally only set them to true if we know it will be necessary, especially `composed`. If we need to maintain to avoid breakage, that's fine.

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,150 @@
...
+
+	_isValidNumber(input) {
+		return input >= 1 && input <= this.maxPageNumber && input !== this.pageNumber;
+	}
+
+	_onNextClicked() {
+		this._dispatchPageChangeEvent(this.pageNumber + 1);
+	}
+
+	_onPageCounterChanged(e) {
```
dbatiste, line 126: Nit: consider changing this to `_onPageSizeChanged` as I think it makes it easier for future readers to understand. 😄

dbatiste (commented):
I haven't reviewed the tests yet, but I've added comments for a few things to address/consider. 😄

src/components/pagination/README.md
```diff
@@ -0,0 +1,29 @@
...
+    import '@brightspace-ui-labs/pagination/pagination.js';
+</script>
+<d2l-labs-pagination></d2l-labs-pagination>
+```
+
+## Properties
+
+| Property | Type | Description |
+|--|--|--|
+| `page-number` | Number (required), default: 1 | The current page number
```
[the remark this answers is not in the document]
ChrisLabattD2L: Yeah, I'd initially not included the `required`, but a teammate had suggested adding it since the initial documentation included it. Since I'm including the default value, I'll remove the `required` from there

src/components/pagination/pagination.js
```diff
@@ -0,0 +1,150 @@
...
+
+		this._dispatchPageChangeEvent(Number(e.target.value));
+	}
+
+	_onPreviousClicked() {
+		this._dispatchPageChangeEvent(this.pageNumber - 1);
+	}
+}
+
+customElements.define('d2l-labs-pagination', Pagination);
```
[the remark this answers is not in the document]
ChrisLabattD2L: Yeah this was mostly to maintain existing usage of the component since it's being used in many places. I don't mind changing the name of it, since I'll have to make various other changes to where it's being used anyways, like having to update the names of the events. It'll be less work to do in the future 👍

src/components/pagination/pagination.js
ChrisLabattD2L: We talked about it quickly over Slack, but I'll mention it here as well in case others see this. I'm deciding to not make this change yet, since I don't know how it'll impact the current uses of the component once it's changed. So the plan is to update everywhere first, then return to here and update to see if anywhere is impacted or not.

src/components/pagination/pager-numeric.js
```diff
@@ -0,0 +1,159 @@
...
+import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
+import { selectStyles } from '@brightspace-ui/core/components/inputs/input-select-styles.js';
+
+class PagerNumeric extends RtlMixin(LocalizeLabsElement(LitElement)) {
+
+	static get properties() {
+		return {
+			maxPageNumber: { type: Number, attribute: 'max-page-number', reflect: true },
+			pageNumber: { type: Number, attribute: 'page-number', reflect: true },
+			pageSize : { type: Number, attribute: 'page-size', reflect: true },
```
dbatiste, line 16: It looks like there is nothing enforcing the `pageSize` is a value in the array. If it's not valid, we could consider using a [custom dependency validator](https://github.com/BrightspaceUI/core/tree/main/mixins/property-required#dependent-properties) to catch it.

That scenario is maybe just an incorrect configuration that wouldn't cause any issues, so not a big deal at this point in time. Don't feel obligated to do that.

test/components/pagination/pagination.test.js
```diff
@@ -0,0 +1,78 @@
...
+import { runConstructor } from '@brightspace-ui/core/tools/constructor-test-helper.js';
+
+const custompageSizes = [2, 5, 37, 159];
+
+async function createComponent({ pageNumber, maxPageNumber, showPageSizeSelector = false, pageSizes, pageSize } = {}) {
+	return await fixture(html`<d2l-labs-pager-numeric
+		page-number="${ifDefined(pageNumber)}"
+		max-page-number="${ifDefined(maxPageNumber)}"
+		?show-page-size-selector="${showPageSizeSelector}"
+		.pageSizes="${ifDefined(pageSizes)}"
```
dbatiste, line 13: Nit: I think `ifDefined` isn't really valid for properties, since it's an attribute directive. I think Lit might tolerate it though. I suspect we can just remove the `ifDefined(...)` here if we're using the property. Or, if we prefer to use the attribute, then we could use `ifDefined(pageSizes ? JSON.stringify(pageSizes) : undefined)`.

dbatiste (approved):
Looks good.

src/components/pagination/pager-numeric.js
ChrisLabattD2L: I think I'm going to omit this validation for the time being. When I tried implementing it, it kept throwing errors because neither `pageSize` nor `pageSizes` are Strings, and `only String properties can be required` according to the error that was being thrown.

PR #360, opened 2024-07-23
Split out axe tests for better tracking

devpow112:
I'm hoping eventually to find a better way to do this so we can track at the test level but based on the current test reporting implementation this is the easiest way to track them separately from just standard old unit tests.


devpow112:
Forgot to update the test reporting config. Will do that now.

dlockhart (approved):

PR #454, opened 2024-11-26
ARK-639 Migrate Attribute Picker

ChrisLabattD2L:
Context for [ARK-639](https://desire2learn.atlassian.net/browse/ARK-639): Migrate Multi-Select to the Labs Monorepo

The purpose of this PR is to move over the code from the [`multi-select`](https://github.com/BrightspaceUILabs/multi-select), with the eventual goal of getting rid of that repo. Note that unlike other such cases, we are not moving everything from that repo, as the rest of the components have been deprecated in favour of the [`d2l-tag-list`](https://daylight.d2l.dev/components/tag-list/#d2l-tag-list) core component. Initially this component used one of those components, but instead of moving that one over, that inner component was heavily simplified to just be the bare bones of what this component needed. The original plan was to reuse the tag-list component, but after facing some difficulties with it, we decided to leave it as-is for now, with the longer term goal of replacing it, though there have been talks about this eventually becoming a core component, so it might also just wait until that.

While moving things over, I also took the time to:
- Simplify, reuse and improve code where possible
- Switch from just using unit tests to also have vdiff tests
  - Many of the existing unit tests were turned into vdiff tests
- Updated lang term names to match the format used in this repo

I was also able to fix up the following defects:
- Deleting an attribute using the `delete` key didn't update screen reader users regarding which element they're on (after the delete)
- Margins weren't configured properly in `rtl` languages, causing some issues with spacing
- When navigating through options in the dropdown, if you far enough that it requires things to scroll, then it would focus on where your mouse is, instead of on the new item
- The scrollbar in the dropdown only appears when there are enough items, instead of always appearing
- Long words that needed to be truncated and appear in the tooltip used to not break up long words, now they do
- The filtering on the dropdown doesn't care about spaces in front of or behind words

I've added the component to the `package.json` `exports` section, so that existing uses of the attribute picker can be switched over to using this.

## Larger Changes

- Updated event names to fit the expected schema:
  - `d2l-attribute-limit-reached` -> `d2l-labs-attribute-picker-limit-reached`
  - `d2l-attributes-changed` -> `d2l-labs-attribute-picker-attributes-changed`

[ARK-639]: https://desire2learn.atlassian.net/browse/ARK-639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/components/attribute-picker/attribute-picker-item.axe.js
```diff
@@ -0,0 +1,9 @@
+import '../../../src/components/attribute-picker/attribute-picker-item.js';
+import { expect, fixture, html } from '@brightspace-ui/testing';
+
+describe('attribute-picker-item', () => {
+	it('should pass all axe tests', async() => {
+		const component = await fixture(html`<d2l-labs-attribute-picker-item text="This is an attribute"></d2l-labs-attribute-picker-item>`);
+		await expect(component).to.be.accessible();
+	});
+});
```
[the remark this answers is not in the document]
dlockhart: It's for the test reporting stuff so that they can be properly categorized as "accessibility tests".

test/components/attribute-picker/attribute-picker-item.test.js
```diff
@@ -0,0 +1,19 @@
...
+			runConstructor('d2l-labs-attribute-picker-item');
+		});
+	});
+
+	describe('eventing', () => {
+		it('should fire d2l-labs-attribute-picker-item-delete when delete button is clicked', async() => {
+			const component = await fixture(html`<d2l-labs-attribute-picker-item text="attribute" deletable></d2l-labs-attribute-picker-item>`);
+			const listener = oneEvent(component, 'd2l-labs-attribute-picker-item-deleted');
+			component.shadowRoot.querySelector('d2l-icon').click();
+			await listener;
```
dlockhart, line 16: Not sure how much you want to fix vs. just migrate over as-is, but the proper way to do this would be:

```suggestion
			clickElem(component.shadowRoot.querySelector('d2l-icon'));
			await oneEvent(component, 'd2l-labs-attribute-picker-item-deleted');
```

test/components/attribute-picker/attribute-picker-item.vdiff.js
```diff
@@ -0,0 +1,61 @@
...
+	});
+
+	it('deletable', async() => {
+		const component = await fixture(html`<d2l-labs-attribute-picker-item text="attribute" deletable></d2l-labs-attribute-picker-item>`);
+		await expect(component).to.be.golden();
+	});
+
+	it('focused', async() => {
+		const component = await fixture(html`<d2l-labs-attribute-picker-item text="attribute"></d2l-labs-attribute-picker-item>`);
+		component.focus();
```
dlockhart, line 17:
```suggestion
		await focusElem(component);
```

test/components/attribute-picker/attribute-picker-item.vdiff.js
```diff
@@ -0,0 +1,61 @@
...
+		await expect(component).to.be.golden();
+	});
+
+	it('focused and deletable', async() => {
+		const component = await fixture(html`<d2l-labs-attribute-picker-item text="attribute" deletable></d2l-labs-attribute-picker-item>`);
+		component.focus();
```
dlockhart, line 23:
```suggestion
		await focusElem(component);
```

test/components/attribute-picker/attribute-picker-item.vdiff.js
```diff
@@ -0,0 +1,61 @@
+import '../../../src/components/attribute-picker/attribute-picker-item.js';
+import { expect, fixture, hoverElem, html, waitUntil } from '@brightspace-ui/testing';
+
+describe('attribute-picker-item', () => {
+	it('default', async() => {
+		const component = await fixture(html`<d2l-labs-attribute-picker-item text="attribute"></d2l-labs-attribute-picker-item>`);
```
dlockhart, line 6: Nit: many of these use the same fixture -- might want to pull its definition out into a `defaultFixture` variable at the top of the file or something.

test/components/attribute-picker/attribute-picker-item.vdiff.js
```diff
@@ -0,0 +1,61 @@
...
+	it('unfocused item with hovered icon', async() => {
+		const component = await fixture(html`<d2l-labs-attribute-picker-item text="attribute" deletable></d2l-labs-attribute-picker-item>`);
+		const icon = component.shadowRoot.querySelector('d2l-icon');
+		await hoverElem(icon);
+		await expect(component).to.be.golden();
+	});
+
+	it('focused item with hovered icon', async() => {
+		const component = await fixture(html`<d2l-labs-attribute-picker-item text="attribute" deletable></d2l-labs-attribute-picker-item>`);
+		component.focus();
```
dlockhart, line 42:
```suggestion
		await focusElem(component);
```

test/components/attribute-picker/attribute-picker-item.vdiff.js
```diff
@@ -0,0 +1,61 @@
...
+		const component = await fixture(html`<d2l-labs-attribute-picker-item text="This is a very long attribute that's being used to show off truncation"></d2l-labs-attribute-picker-item>`);
+		await expect(component).to.be.golden();
+	});
+
+	it('truncated with tooltip', async() => {
+		const component = await fixture(html`<d2l-labs-attribute-picker-item style="margin-top: 50px;" text="This is a very long attribute that's being used to show off truncation"></d2l-labs-attribute-picker-item>`);
+		const tooltip = component.shadowRoot.querySelector('d2l-tooltip');
+
+		await hoverElem(component);
+		await waitUntil(() => tooltip._truncating);
```
dlockhart, line 58: Hopefully this could prevent poking into the internals like this...
```suggestion
		hoverElem(component);
		await oneEvent(component, 'd2l-tooltip-show');
```

test/components/attribute-picker/attribute-picker.test.js
```diff
@@ -0,0 +1,149 @@
...
+		});
+	});
+
+	describe('interaction', () => {
+		it('should prevent unlisted if allow-freeform is disabled.', async() => {
+			const component = await createComponent(attributeList, { allowFreeform: false });
+			const input = component.shadowRoot.querySelector('input');
+			input.focus();
+
+			component._text = 'unlisted attribute';
```
dlockhart, line 47: Might be worth trying to use `sendKeysElem` instead of focusing and setting the text programatically.

test/components/attribute-picker/attribute-picker-item.test.js
ChrisLabattD2L: I don't mind making the switch over to all of those!

test/components/attribute-picker/attribute-picker-item.vdiff.js
ChrisLabattD2L: Didn't realize they could not be awaited, so I'll have to try that more with other things that failed for me!

src/components/attribute-picker/attribute-picker.js
```diff
@@ -0,0 +1,446 @@
...
+
+	// Returns true or false depending on if the attribute was successfully added. Fires the d2l-attribute-limit-reached event, if necessary.
+	async addAttribute(newAttribute) {
+		if (!newAttribute || this.attributeList.findIndex(attribute => attribute.name === newAttribute.name) >= 0) return false;
+
+		if (this.limit && this.attributeList.length >= this.limit) {
+			this.dispatchEvent(new CustomEvent('d2l-labs-attribute-picker-limit-reached', {
+				bubbles: true, composed: true, detail: { limit: this.limit }
+			}));
+			return false;
```
BhamelinD2L, line 267: There doesn't seem to be a tooltip stating the limit has been reached in this version. Did we intend to get rid of it?
image

ChrisLabattD2L: That tooltip is found in the [foundation-components part](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/main/features/discover/d2l-discover-attribute-picker.js#L52) that implements this

BhamelinD2L: Should it be, or is this something that should belong to the picker itself for other consumers to benefit from?

test/components/attribute-picker/attribute-picker-item.test.js
ChrisLabattD2L: Okay, so I updated the tests to use the helpers where possible, with the exception of [here](https://github.com/BrightspaceUI/labs/blob/clabatt/ARK-639-Migrate-Attribute-Picker/test/components/attribute-picker/attribute-picker.vdiff.js#L121) where `clickElem` doesn't seem to get registered (regardless of if it's awaited or not). I also realized that there's no helper for `blur`, but it's not the end of the world.

test/components/attribute-picker/attribute-picker-item.vdiff.js
ChrisLabattD2L: Don't think I've ever used or heard of `defaultFixture` before, so I'll have to read more about it

test/components/attribute-picker/attribute-picker-item.test.js
dlockhart: Hmm, I don't love that `<d2l-icon role="button">` business since it's not going to be keyboard accessible. Would converting it to `<d2l-button-icon>` be terrible?

Yeah for blur typically I'll just do `clickAt(0, 0)`.

test/components/attribute-picker/attribute-picker-item.vdiff.js
dlockhart: Haha yeah exactly. Just like:

```javascript
const defaultFixture = html`<d2l-labs->...`;
```

ChrisLabattD2L: I created it as a function in the end, similar to what I'd done with the attribute picker test, so that it could also be used for the deletable state, while also resetting between tests 👍

test/components/attribute-picker/attribute-picker-item.test.js
ChrisLabattD2L: So, we had to do some magic a while ago, but it's actually keyboard accessible, just not in the traditional way, partially because of how things with the `application` role function. Hmm, I can look into using the button instead, I think we had issues with it initially because of it's sizing spacing compared to item. I can revisit that on Monday and talk with my team about making that change

src/components/attribute-picker/attribute-picker.js
ChrisLabattD2L: Hmm, that's a good question, we can talk more about it during stand-up on Monday, along with the point Dave just made about using the icon button instead of just the icon. I probably wouldn't hurt to have the tooltip as part of it.

test/components/attribute-picker/attribute-picker-item.test.js
ChrisLabattD2L: So after messing around with the button, there were quire a few complications with using the button icon instead. The main ones where the size of the icon, which didn't seem like we could adjust in the way we wanted, it broke the navigation between the items, and also we couldn't show the hover state when the item wasn't selected, meaning they wouldn't know they could delete the item.

I do think the current solution is essentially fully accessible with the main potential exceptions being if the hover state of the icons is different enough. Would you be okay with things staying as they are currently since it's a labs component? And then if it does look like it'll be promoted to something bigger, than we make those improvements then?

dlockhart: Thanks for looking into it. Yeah that's totally fine, although possibly a good middle ground solution would just be to use `<button aria-label="whatever"><d2l-icon ...>` and then style the button & icon however you want. Then you'll get the right keyboard/focus accessibility for free.

ChrisLabattD2L: While working on that I made a realization that I was being silly with the `d2l-button-icon`, I'd used the same icon (`close-large-thick`) instead of switching to `close-small`, which fits a lot more appropriately. So I'm back to trying to get that to work!

ChrisLabattD2L: Had to make a few modifications, but was able to switch to using `d2l-button-icon` along with make a few other improvements, so it should be good to go now 😄

src/components/attribute-picker/attribute-picker-item.js
```diff
@@ -29,39 +30,35 @@ class AttributePickerItem extends LocalizeLabsElement(LitElement) {
...
+				gap: 0.2rem;
 				padding: 0.25rem 0.75rem 0.2rem;
 				user-select: none;
 			}
 			:host([deletable]) .d2l-labs-attribute-picker-item-wrapper {
-				padding: 0.25rem 0.6rem 0.25rem 0.55rem;
+				padding: 0.1rem 0.1rem 0.1rem 0.55rem;
+			}
+			:host([dir="rtl"][deletable]) .d2l-labs-attribute-picker-item-wrapper {
+				padding: 0.1rem 0.55rem 0.1rem 0.1rem;
```
dlockhart, line 50: Optional: We've actually been slowly switching over to [CSS logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values) and removing `RtlMixin`. Makes things a lot cleaner.

src/components/attribute-picker/attribute-picker-item.js
```diff
@@ -88,7 +97,7 @@ class AttributePickerItem extends LocalizeLabsElement(LitElement) {
 		return html`
 			<div class="d2l-labs-attribute-picker-item-wrapper">
 				<div id="d2l-labs-attribute-picker-item-text" class="d2l-body-compact d2l-labs-attribute-picker-item-text">${this.text}</div>
-				<d2l-icon aria-label="${this.localize('components:attributePicker:removeValue', 'value', this.text)}" role="button" icon="d2l-tier1:close-large-thick" ?hidden="${!this.deletable}" @click="${this._onDeleteItem}"></d2l-icon>
+				<d2l-button-icon role="button" tabindex="-1" text="${this.localize('components:attributePicker:removeValue', 'value', this.text)}" icon="tier1:close-small" ?hidden="${!this.deletable}" @click="${this._onDeleteItem}"></d2l-button-icon>
```
dlockhart, line 100: Hopefully `role` and `tabindex` aren't needed? `<d2l-button-icon>` isn't actually a `button`, it just wraps a native `<button>` -- so that's the element that has the button role and receives focus.

src/components/attribute-picker/attribute-picker.js
```diff
@@ -41,6 +38,9 @@ class AttributePicker extends ArrowKeysMixin(RtlMixin(LocalizeLabsElement(LitEle
 			*/
 			invalidTooltipText: { type: String, attribute: 'invalid-tooltip-text', reflect: true },
 
+			/* Required. The label associated with the attribute picker for screen reader users */
+			label: { type: String, reflect: true },
```
dlockhart, line 42: Nice. This will help axe be less upset!

src/components/attribute-picker/attribute-picker.js
```diff
@@ -62,6 +62,9 @@ class AttributePicker extends ArrowKeysMixin(RtlMixin(LocalizeLabsElement(LitEle
 			/* When true, the user currently has focus within the input */
 			_inputFocused: { type: Boolean, reflect: false },
 
+			/* When true, the user has reached the limit of attributes that can be added */
+			_limitReached: { type: Boolean, reflect: false },
```
dlockhart, line 66: Could use reactive state for this kind of thing.

```suggestion
			_limitReached: { state: true },
```

src/components/attribute-picker/attribute-picker.js
```diff
@@ -167,6 +173,12 @@ class AttributePicker extends ArrowKeysMixin(RtlMixin(LocalizeLabsElement(LitEle
 		this._inputFocused = false;
 	}
 
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		this.addEventListener('focusin', this._onFocusIn);
+		this.addEventListener('focusout', this._onFocusOut);
```
dlockhart, line 179: The recommended approach for this kind of thing is to wire event listeners up in `connectedCallback` and remove them in `disconnectedCallback`. Otherwise if this element gets removed from the DOM the listeners are still there, and then if it gets re-added it'll add a second set of listeners.

Otherwise you could keep the listeners on the `<div>` and then Lit will manage them for you.

ChrisLabattD2L: I'd put them in `firstUpdated` based on how the [`tag-list-item-mixin`](https://github.com/BrightspaceUI/core/blob/main/components/tag-list/tag-list-item-mixin.js#L187) did it, but moving them to connected & disconnected also works, so I'll make that switch! I'd initially had it tied to a div, but it had problems keeping the focus when switching between items (even though they were inside), so moving it to here resolved the issue.

src/components/attribute-picker/attribute-picker-item.js
ChrisLabattD2L: I can probably remove the role, but tabindex is actually needed (similar to the [`tag-list-item-mixin`](https://github.com/BrightspaceUI/core/blob/main/components/tag-list/tag-list-item-mixin.js#L309)), otherwise I was noticing some weird behaviour with tabbing, especially when trying to tab backwards.

src/components/attribute-picker/attribute-picker.js
dlockhart: Clearly nobody's perfect haha

src/components/attribute-picker/attribute-picker-item.js
```diff
@@ -4,22 +4,13 @@ import '@brightspace-ui/core/components/tooltip/tooltip.js';
...
-			},
-			tabIndex: {
-				type: Number,
-				reflect: true
-			},
-			text: {
-				type: String
-			}
+			deletable: { type: Boolean, reflect: true },
+			tabIndex: { type: Number, reflect: true },
```
dlockhart, line 12: `tabIndex` is a native thing that all elements have, so you shouldn't need to redefine it as a Lit reactive property _unless_ you need the act of changing it to trigger a re-render.

ChrisLabattD2L: Without it in properties, the `runConstructor` actually fails, which is why I'd initially done it, but I move the initialization of it to `firstUpdated`, then it actually works as expected, so I'll make that change 👍

dlockhart (approved):
This is looking good to me!

src/components/attribute-picker/attribute-picker.js
margaree: We do have this [documented](https://daylight.d2l.dev/developing/best-practices/events/#originating-from-the-host-element) in Daylight as recommending using `firstUpdated` for cases like this so if our recommendation is changing here we should keep that consistent.

BhamelinD2L (approved):
Looks great, thanks so much!

src/components/attribute-picker/attribute-picker.js
dlockhart: Unresolved this thread, as Margaree is correct that the way you had it initially is the correct approach.

dlockhart (approved):

src/components/attribute-picker/attribute-picker.js
ChrisLabattD2L: No problem! At least we know what the official recommendation is now 😄

dlockhart: Haha I'm pretty sure I've made this mixup a couple times, hence the doc that got written. Maybe this time it'll stick!

PR #495, opened 2025-01-09
feat: migrate card-overlay

GZolla:
[JIRA](https://desire2learn.atlassian.net/browse/GAUD-7361)


src/components/card-overlay/card-overlay.js
dlockhart, line 1: This file will need to be exported from `package.json`.

dlockhart (approved):

PR #477, opened 2024-12-19
feat: Add height variable

ataromoku:
## Jira ticket:
https://desire2learn.atlassian.net/browse/PHNX-2096

The `height: 100%` on the flyout in Lessons causes the page to shift upwards when anchor links are used (`element.scrollIntoView`). Changing the height to `auto` resolves this issue.

### Related PRs:
* https://github.com/Brightspace/lms/pull/61106

PR #506, opened 2025-01-22
fix: Opt-out dialog is vertically cut off if height is set to auto

ataromoku:
## Jira ticket
https://desire2learn.atlassian.net/browse/PHNX-2526

## Problem
Previous [PR1](https://github.com/BrightspaceUI/labs/pull/477) [PR2](https://github.com/Brightspace/lms/pull/61106) addressed an issue where the entire page shifted upward when `scrollIntoView` was called. This happened because the `opt-in-flayout` component overflowed out of view. However, that fix introduced a new problem, the dialog window is now getting cut off because the component's height is too small to fit it in.

## Solution
Set the height to `auto` by default and use `100%` to fill the entire space only when the `opt-in-flyout` is in *open* state. Hide the content behind the `opt-in-flyout` using a fade.

---
### Demo

<details>
  <summary>Normal style</summary>

https://github.com/user-attachments/assets/e8bd3e53-4b47-4e21-b113-623dab5f3a2f

</details>

<details>
  <summary>Position fixed style</summary>

https://github.com/user-attachments/assets/17a70b06-8fce-4f8c-bd7b-921fa333287d

</details>

<details>
  <summary>Before</summary>

https://github.com/user-attachments/assets/4f268f83-27fe-46bf-a346-de7b6ce6e304

</details>


src/components/opt-in-flyout/flyout-impl.js
```diff
@@ -47,13 +47,19 @@ class FlyoutImplementation extends composeMixins(
...
-					height: var(--d2l-flyout-custom-element-height, 100%);
+					height: var(--d2l-flyout-custom-element-height, auto);
 					overflow: hidden;
 					pointer-events: none;
 					position: absolute;
 					width: 100%;
 				}
 
+				:host([opened]) {
+					background-color: rgba(255, 255, 255, 0.7);
```
dbatiste, line 58: Curious what the semi-transparency is being used for here.

dbatiste: I see what this is being used for now. Follow-up question, do you know why this is using `white` as opposed to regolith like [d2l-backdrop](https://github.com/BrightspaceUI/core/blob/main/components/backdrop/backdrop.js#L46). I see the transparency is at least consistent.

PR #523, opened 2025-02-04
feat: Switch to using proxies for reactive store consumers

jorge-ramirez-arredondo:
This change makes it so the `StoreConsumer` and the context `Provider`/`Consumer` pair are instantiated as proxies for the store they're connecting to.

This makes it possible for all methods and properties defined within the store (not just ones defined in the static `properties`) to be accessed and used from any consumer.

Did this work as part of this refactor: https://desire2learn.atlassian.net/browse/MED-1354


dlockhart (approved):
Cool!

PR #542, opened 2025-02-25
ARK-1769 Force input to have inherited font

ChrisLabattD2L:
It's been pointed out to me that the attribute-picker's text wasn't using the d2l-input style's, making it smaller than it should be, but by making the `font-family` and `font-size` inherited, it resolves the issues.

Before:

image

After:

image


dbatiste (approved):

dlockhart:
The `/jira` command doesn't work for public repos.

PR #549, opened 2025-03-04
MED-1431 Make sure the ReactiveStore supports reactive property inheritance

jorge-ramirez-arredondo:
This fixes an issue with the original implementation that made it so extending a previously defined ReactiveStore and implementing new reactive properties would override any previous reactive properties. With this fix, properties inherited from ancestors will be implemented before descendants.


dlockhart (approved):
Nice!

PR #561, opened 2025-03-13
feat: empty commit

GZolla:
[JIRA](https://desire2learn.atlassian.net/browse/GAUD-7621)

Empty PR to test versioning releases.
I noticed it released a major version earlier today(though it could not publish to npm). I've looked through the source code and apparently it looks for commits after the highest versioning tag(from what I gathered no github release needs to be attached to the tag). I added a tag after the merge, hopefully that resolves the issue going forward.

FYI, I expect this commit to fail in publishing to npm since npm won't allow publishing 24 hours after unpublishing.


dlockhart (approved):
Makes sense. If we port history over again, I guess we'll have to comb through it looking for major version messages and manually remove them from the commit messages. 

Probably not worth porting the history.

PR #562, opened 2025-03-13
feat: add lit-router [skip ci]

GZolla:
[JIRA](https://desire2learn.atlassian.net/browse/GAUD-7626)

I think an easy way to prevent the undesires release is to disable the workflow -> merge this PR -> add tags -> enable workflow -> merge empty PR. I still feel the history is valuable specially for this component that has recent changes(hence why I looked into preserving history in the first place).

Otherwise, I'll make a separate PR without the history.


NicholasHallman:
Thanks for moving this over and working hard to preserve the commit history!

bearfriend:
This is what I was thinking the other day, but using `[skip ci]` instead of disabling the workflow. That worked when I force pushed to remove the breaking change release commit.

dlockhart (approved):
LGTM!

PR #565, opened 2025-03-14
fix: import lang terms

GZolla:
[JIRA](https://desire2learn.atlassian.net/browse/GAUD-7102)


bearfriend (approved):
Did we let Localization team know about these?

GZolla:
> Did we let Localization team know about these?

Yep, I'll send her the PR too.

PR #567, opened 2025-03-17
feat: Add text changed event to d2l-labs-attribute-picker (MED-1434)

MarquisLP:
This PR updates the `d2l-labs-attribute-picker` to dispatch a new custom event named `d2l-labs-attribute-picker-text-changed` whenever the attribute picker's text input value is changed. The event contains the updated text input value.

This allows consumers of `d2l-labs-attribute-picker` to track and respond to user input in real time. For example, this can be used to send a data query to the backend each time the user types a new value. The fetched data can then be used to update the list of available attributes displayed in the dropdown.

Example - user search:
Screenshot 2025-03-17 at 11 53 54 AM


src/components/attribute-picker/attribute-picker.js
```diff
@@ -426,10 +426,18 @@ class AttributePicker extends ArrowKeysMixin(LocalizeLabsElement(LitElement)) {
 	}
 
 	_onInputTextChanged(e) {
+		e.stopPropagation();
```
MarquisLP, line 429: Right now, the `input` event from the inner `<input>` element bubbles up through the DOM and can be caught by consumers of `<d2l-labs-attribute-picker>`. This event has `target` set to `<d2l-labs-attribute-picker>`. Since `<d2l-labs-attribute-picker>` doesn't have a `value` property, `event.target.value` ends up being `undefined`.

As such, to discourage listening to the `input` event and encourage listening to `d2l-attribute-picker-text-changed` instead, the `input` event's propagation is stopped here.

src/components/attribute-picker/attribute-picker.js
```diff
@@ -426,10 +426,18 @@ class AttributePicker extends ArrowKeysMixin(LocalizeLabsElement(LitElement)) {
...
+
 		this._text = e.target.value;
 		if (this._dropdownIndex >= 0) {
 			this.allowFreeform ? this._dropdownIndex = -1 : this._dropdownIndex = 0;
 		}
+
+		this.dispatchEvent(new CustomEvent('d2l-labs-attribute-picker-text-changed', {
+			detail: { text: this._text },
+			bubbles: true,
+			composed: true,
```
dlockhart, line 439: In general, I typically recommend _not_ bubbling up through the DOM or being composed unless you absolutely need it. It's just nicer to not having events flying around past nodes that will never need to care about them. That would still allow a consumer to wire up a listener directly to the attribute picker, which is likely your use case here.

src/components/attribute-picker/attribute-picker.js
dlockhart: If you suppress the `input` event, existing consumers [like this one](https://github.com/Brightspace/manager-view-fra/blob/2f258ffa9a50820b23312135455aa65ef089ecd0/src/group/mr-group-membership-rules.js#L261) are going to stop working I think.

src/components/attribute-picker/attribute-picker.js
MarquisLP: Gotcha, that makes sense. I had enabled `bubbles` and `composed` for consistency with the other two event types in this component, but it's true that our use-case has no need for the event to bubble up further than 1 level. I've tested this change just now and it works fine, so I'll push a commit shortly.

src/components/attribute-picker/attribute-picker.js
MarquisLP: I see, thanks for catching that. I'll remove this line then to avoid breakage. I suppose existing consumers can make the decision to switch over to the new event as they see fit.

test/components/attribute-picker/attribute-picker.test.js
```diff
@@ -145,5 +145,28 @@ describe('attribute-picker', () => {
 			expect(detail.limit).to.equal(5);
 			expect(component.attributeList.length).to.equal(5);
 		});
+
+		it('should fire the d2l-labs-attribute-picker-text-changed event when inputting text into the text input', async() => {
+			const component = await createComponent(attributeList, { limit: 5 });
+
+			const input = component.shadowRoot.querySelector('input');
+
+			await focusElem(input);
```
dlockhart, line 154: Not a big deal, but `sendKeysElem` will automatically focus for you, so this line shouldn't be necessary.

test/components/attribute-picker/attribute-picker.test.js
```diff
@@ -145,5 +145,28 @@ describe('attribute-picker', () => {
...
+
+			const listener1 = oneEvent(component, 'd2l-labs-attribute-picker-text-changed');
+			await sendKeysElem(input, 'type', 'a');
+			const { detail: detail1 } = await listener1;
+			expect(detail1.text).to.equal('a');
```
dlockhart, line 159: Very nit, but to avoid having to separate out the `oneEvent` listener you should be able to exclude the `await` on `sendKeysElem`. Like this:

```suggestion
			sendKeysElem(input, 'type', 'a');
			const { detail: detail1 } = await oneEvent(component, 'd2l-labs-attribute-picker-text-changed');
			expect(detail1.text).to.equal('a');
```

dlockhart (approved):

dlockhart (approved):

PR #581, opened 2025-03-24
feat: migrate media-player [skip ci]

GZolla:
[JIRA](https://desire2learn.atlassian.net/browse/GAUD-7678)


test/components/media-player/.eslintrc.json
dlockhart, line 1: Remove?

dlockhart (commented):
Missing `package.json` entry. Also this might be a good component to explicitly list the `Media` team in the CODEOWNERS for it specifically? They tend to still make all the changes to it.

dlockhart (approved):

PR #589, opened 2025-03-28
fix: accessibility issues with navigation

GZolla:
[JIRA](https://desire2learn.atlassian.net/browse/GAUD-7686)

Adding the void `href` on navigation-skip makes the anchor dispatch click events for keyboard. It also announces the anchor as a link which it was not doing before. 

[JIRA](https://desire2learn.atlassian.net/browse/GAUD-7688)

Adding label properties to add them to the underlying button instead of the custom opener component. NVDA does not seem to like the custom component having labels or ids(no idea why the latter is the case).

### DIFF
--- a/src/components/navigation/navigation-dropdown-button-custom.js
+++ b/src/components/navigation/navigation-dropdown-button-custom.js
@@ -7,8 +7,7 @@ class NavigationDropdownButtonCustom extends DropdownOpenerMixin(LitElement) {
 
 	static get properties() {
 		return {
-			openerLabel: { type: String, attribute: 'opener-label' },
-			openerLabelledby: { type: String, attribute: 'opener-labelledby' }
+			openerLabel: { type: String, attribute: 'opener-label' }
 		};
 	}
 
@@ -30,8 +29,7 @@ class NavigationDropdownButtonCustom extends DropdownOpenerMixin(LitElement) {
 			<button
 				type="button"
 				aria-haspopup="menu"
-				aria-label="${ifDefined(this.openerLabel)}"
-				aria-labelledby="${ifDefined(this.openerLabelledby)}">
+				aria-label="${ifDefined(this.openerLabel)}">
 				<span class="d2l-labs-navigation-highlight-border"></span>
 				<slot name="opener"></slot>
 			</button>
--- a/src/components/navigation/navigation-dropdown-button-custom.js
+++ b/src/components/navigation/navigation-dropdown-button-custom.js
@@ -1,8 +1,16 @@
 import { css, html, LitElement } from 'lit';
 import { highlightBorderStyles, highlightButtonStyles } from './navigation-styles.js';
 import { DropdownOpenerMixin } from '@brightspace-ui/core/components/dropdown/dropdown-opener-mixin.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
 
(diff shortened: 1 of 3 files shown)


src/components/navigation/navigation-dropdown-button-custom.js
```diff
@@ -19,7 +27,11 @@ class NavigationDropdownButtonCustom extends DropdownOpenerMixin(LitElement) {
 
 	render() {
 		return html`
-			<button type="button">
+			<button
+				type="button"
+				aria-haspopup="menu"
+				aria-label="${ifDefined(this.openerLabel)}"
+				aria-labelledby="${ifDefined(this.openerLabelledby)}">
```
bearfriend, line 34: Does `aria-leballedby` work with elements in the slot? I know it can't cross the shadow-light boundary, but I hadn't considered slotted content.

bearfriend: Maybe I'm confused about what is labelling this button.

src/components/navigation/navigation-dropdown-button-custom.js
```diff
@@ -1,9 +1,17 @@
...
 import { highlightBorderStyles, highlightButtonStyles } from './navigation-styles.js';
 import { DropdownOpenerMixin } from '@brightspace-ui/core/components/dropdown/dropdown-opener-mixin.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
 
 class NavigationDropdownButtonCustom extends DropdownOpenerMixin(LitElement) {
 
+	static get properties() {
+		return {
+			openerLabel: { type: String, attribute: 'opener-label' },
+			openerLabelledby: { type: String, attribute: 'opener-labelledby' }
```
dlockhart, line 11: I don't think `labelled-by` is going to work here since the elements it references can't span shadow roots.

src/components/navigation/navigation-skip.js
```diff
@@ -44,14 +44,7 @@ class NavigationSkip extends FocusMixin(PropertyRequiredMixin(LitElement)) {
...
 	render() {
-		return html`<a tabindex="0" @keydown="${this._handleKeyDown}" class="vdiff-target">${this.text}</a>`;
-	}
-
-	_handleKeyDown(e) {
-		if (e.keyCode === 13) {
-			e.preventDefault();
-			this.dispatchEvent(new CustomEvent('click', { bubbles: true, composed: true }));
-		}
+		return html`<a href="javascript:void(0);" class="vdiff-target">${this.text}</a>`;
```
dlockhart, line 47: This is what I noticed too. This might be a good place for a short comment explaining why the `href` is needed, seeing as it's such a weird situation.

src/components/navigation/navigation-dropdown-button-custom.js
dlockhart: Yeah I had a similar comment (deleted now).

GZolla: I think you are right, as Dave also mentions, I tested it out and it does break.

GZolla: the solution for the navigation dropdown defect would likely use label anyway since the hidden span is not being read currently.

bearfriend (approved):

PR #599, opened 2025-04-10
fix: replace usages of checkbox spacer

dlockhart:
[GAUD-7778](https://desire2learn.atlassian.net/browse/GAUD-7778)

A new `supporting` slot has been added to `<d2l-input-checkbox>` to handle use cases where additional components need to be rendered so they visually appear nested (below and indented) inside the checkbox.

Previously, `<d2l-input-checkbox-spacer>` was used to accomplish this, but we'd like to remove that component in order to solve some spacing issues with groups of checkboxes.

[GAUD-7778]: https://desire2learn.atlassian.net/browse/GAUD-7778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Expecting potentially some very minor visual diffs. [Vdiff Report](https://vdiff.d2l.dev/BrightspaceUI/labs/f4d23d4e15ecdd70ce7f7ef81de910d718d8086d/2025-04-10_20-35-25/report/)


src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -64,17 +53,14 @@ class CheckboxDrawer extends LocalizeLabsElement(SkeletonMixin(LitElement)) {
...
+					@change="${this._onCheckboxChange}">
+					${this.label}
+					<div slot="inline-help">${this.description}</div>
+					<d2l-expand-collapse-content slot="supporting" ?expanded="${this.checked}" aria-busy=${this._expandCollapseBusy ? 'true' : 'false'} @d2l-expand-collapse-content-expand="${this._onExpandCollapseContentToggled}" @d2l-expand-collapse-content-collapse="${this._onExpandCollapseContentToggled}">
+						<slot></slot>
+					</d2l-expand-collapse-content>
+				</d2l-input-checkbox>
 			`}
-			<d2l-input-checkbox-spacer class="d2l-input-checkbox-spacer">
-				<div class="d2l-input-checkbox-description d2l-skeletize">${this.description}</div>
```
dlockhart, line 70: This "description" is styled to look like "body small", which is what the `inline-help` slot is for -- so I've moved that in there.

src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -64,17 +53,14 @@ class CheckboxDrawer extends LocalizeLabsElement(SkeletonMixin(LitElement)) {
...
-			</d2l-input-checkbox-spacer>
-			<d2l-input-checkbox-spacer class="d2l-input-checkbox-spacer">
```
dlockhart, line 72: The rest of this can go in the `supporting` slot.

Screenshot 2025-04-10 at 4 14 51 PM

src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -54,27 +56,34 @@ class CheckboxDrawer extends LocalizeLabsElement(SkeletonMixin(LitElement)) {
...
-				<d2l-expand-collapse-content ?expanded="${this.checked}" aria-busy=${this._expandCollapseBusy ? 'true' : 'false'} @d2l-expand-collapse-content-expand="${this._onExpandCollapseContentToggled}" @d2l-expand-collapse-content-collapse="${this._onExpandCollapseContentToggled}">
-					<div class="d2l-checkbox-content-margin"></div>
+				<div class="d2l-input-read-only-spacer">
+					<div class="d2l-input-checkbox-description d2l-skeletize">${this.description}</div>
+				</div>
+				<div class="d2l-input-read-only-spacer">
+					<d2l-expand-collapse-content ?expanded="${this.checked}" aria-busy=${this._expandCollapseBusy ? 'true' : 'false'} @d2l-expand-collapse-content-expand="${this._onExpandCollapseContentToggled}" @d2l-expand-collapse-content-collapse="${this._onExpandCollapseContentToggled}">
+						<div class="d2l-checkbox-content-margin"></div>
```
GZolla, line 68: Is this margin not needed for the checkbox case?

src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -54,27 +56,34 @@ class CheckboxDrawer extends LocalizeLabsElement(SkeletonMixin(LitElement)) {
...
+		return html`
+			<d2l-input-checkbox
+				?checked="${this.checked}"
+				class="d2l-input-checkbox"
+				?skeleton="${this.skeleton}"
+				description=${this.localize(`components:checkboxDrawer:checkbox${this.checked ? 'Expanded' : 'Collapsed'}`)}
+				@change="${this._onCheckboxChange}">
+				${this.label}
+				<div slot="inline-help">${this.description}</div>
+				<d2l-expand-collapse-content slot="supporting" ?expanded="${this.checked}" aria-busy=${this._expandCollapseBusy ? 'true' : 'false'} @d2l-expand-collapse-content-expand="${this._onExpandCollapseContentToggled}" @d2l-expand-collapse-content-collapse="${this._onExpandCollapseContentToggled}">
```
GZolla, line 83: Nit: Could the slot attribute be set conditionally using `ifDefined` and the margin(if not needed) added conditionally too? So you can pull this component out to a reusable variable to avoid repeating code, seems hacky so I'm fine either way.

src/components/checkbox-drawer/checkbox-drawer.js
dlockhart: Good catch and no, it isn't. The `supporting` slot comes with built-in margin between the checkbox and itself.

src/components/checkbox-drawer/checkbox-drawer.js
dlockhart: Possibly, although there would also be the `d2l-input-read-only-spacer` CSS class that would need to be applied in the read-only case as well. In the end, the rendering was so different that it seemed clearer to just totally separate them.

GZolla (approved):

src/components/checkbox-drawer/checkbox-drawer.js
dbatiste: Unfortunate that we need to keep the `color` and `font-size` properties around for the read-only version. 😢

dbatiste (approved):

src/components/checkbox-drawer/checkbox-drawer.js
dlockhart: Yeah this made me a little sad, but also "labs".

PR #616, opened 2025-04-30
jprouse/MAV-3764-update-view-toggle-media-query

JakeProuse:
## JIRA Link
https://desire2learn.atlassian.net/browse/MAV-3764

## Description
In the H5P Direct Access Modal we use the toggle to switch between list modes and noticed that the component disappears at viewports lower than 525px. This PR sets the toggles to stay visible at lower viewports, disables the wrapping of toggle buttons, and adds some vdiff tests for the component.


test/components/view-toggle/view-toggle.vdiff.js
```diff
@@ -0,0 +1,41 @@
...
+
+const basicFixture = html`<d2l-labs-view-toggle
+toggle-options='[{"text":"Bananas","val":"overall"},{"text":"Minions","val":"minios"},{"text":"Pyjamas","val":"subject"}]'
+text="Toggle: "></d2l-labs-view-toggle>`;
+
+const noWrapFixture = html`<d2l-labs-view-toggle
+toggle-options='[{"text":"Bananas","val":"overall"},{"text":"Minions","val":"minios"},{"text":"Pyjamas","val":"subject"},{"text":"Elephant","val":"longmemory"},{"text":"Tiger","val":"fast"}]'
+text="Toggle: "></d2l-labs-view-toggle>`;
+
+const viewports = [
```
JakeProuse, line 12: Noticed a lot of labs tests just pick a mobile and a desktop viewport, but added in the viewports we use in our [own code](https://github.com/BrightspaceUI/creator-plus-authoring-tools/blob/main/test/utilities/constants.js#L6) for some extra coverage.

test/components/view-toggle/view-toggle.vdiff.js
```diff
@@ -0,0 +1,41 @@
...
+	{ viewport: { height: 800, width: 1230 } },
+	{ viewport: { height: 800, width: 865 } },
+	{ viewport: { height: 800, width: 767 } },
+	{ viewport: { height: 800, width: 614 } },
+	{ viewport: { height: 800, width: 430 } }
+];
+
+describe('d2l-labs-view-toggle', () => {
+	it('should not wrap', async() => {
```
JakeProuse, line 21: The snapshot of this test is the same as the mobile viewport one, but this does lock in a test specifically for a too-long set of toggles. Maybe I should update the basicFixture to just be two toggles?

test/components/view-toggle/view-toggle.vdiff.js
dlockhart: Seeing as the component no longer has any responsive styles, I honestly wouldn't test any of these as they're just taking up time & space. I'd keep your "should not wrap" test on a smaller viewport just to assert that things don't wrap, but leave out the rest of the viewport tests.

src/components/view-toggle/view-toggle.js
```diff
@@ -70,23 +70,18 @@ class ViewToggle extends LitElement {
 			button[aria-pressed="true"]:hover, button[aria-pressed="true"]:focus {
 				box-shadow: inset 0 0 0 2px #ffffff;
 			}
+
 			:host {
-				display: flex;
-				width: 100%;
+				display: block;
+				margin: 0 -0.9rem;
```
dlockhart, line 76: I know this was existing before, but I'm pretty confused as to why it would use a negative margin on the host and then a positive inverse one on the container. 😕

svanherk: Yeah, very weird. Only other time I've done something like that was to allow focus styles to overflow, which doesn't really apply here since the focus styles go inwards 🤔

JakeProuse: Yeah wasn't sure what was going on there 😕

JakeProuse: Interesting, removed the margin rules and got some vdiff changes https://github.com/BrightspaceUI/labs/pull/621/files

dlockhart: That is a bit weird, but I think it's OK. If I play around in the demo and toggle between your branch and `main`, the actual width is the same.

One thing that'll need fixing though is the gap between the label and the first button has shrunk to nothing, whereas before I think it was relying on whitespace between the `<span>` and the `<button>`. I think you can play with either a `margin-inline-end` on the `span` or a custom `gap` to remove that change.

JakeProuse:
@dlockhart / @svanherk have you seen 'Expected — Waiting for status to be reported' on an action before? Doesn't give a re-run option and I don't see it having been run in Actions.

dlockhart:
Sometimes I think things get stuck -- try pushing an empty commit.

dlockhart (approved):
Seems good to me! Thanks for fixing this up.

PR #627, opened 2025-05-07
feat: minor release for toggle component

RPena-D2l:
https://desire2learn.atlassian.net/browse/MAV-3764


dlockhart (approved):
Technically this should probably be a `fix:` but I'm also OK calling this a feature haha.

PR #630, opened 2025-05-09
GAUD-7943 - Fix up demo pages

svanherk:
Add consistency, add missing demos to the `index` page, and fix a few pages.


demo/components/accordion/accordion.html
svanherk, line 1: This demo page is so confusing, I added a copy that sets visible labels to make the other demos easier to follow.

demo/utilities/router/index.html
```diff
@@ -2,7 +2,7 @@
 <html>
 	<head>
 		<title>Router Playground</title>
-		<script type="module" src="/example/app.js"></script>
+		<script type="module" src="/demo/utilities/router/app.js"></script>
```
svanherk, line 5: The router demo was broken. It now "works", but is still not great because if you navigate away and press the back button it's completely lost. This is due to how we have the repo setup and how the router inherently works - we'll need to come up with something else here.

GZolla (approved):
I think most of this mess is on me 😞. Thanks for the changes!

svanherk:
> I think most of this mess is on me 😞. Thanks for the changes!

No worries! A lot of this was in the original demos that were copied over.

demo/utilities/router/index.html
dlockhart: The router's going to move out of here into its own repo soonish anyway.

dlockhart (approved):
Thanks!

PR #633, opened 2025-05-12
GAUD-7949 - Hookup live previews

svanherk:
This basically copies what `core` is doing, but `core` has the luxury of getting to use its own styles, demo components, etc. ~In other repos, it creates an ugly `../../../../node_modules/...` path in all the demos, and we need to call out `core` components only used in the demos in `rollup.config.js` so rollup knows to include them. Please let me know if I'm just `rollup`-ing wrong!~
Update: `@web/rollup-plugin-html` gets around this!

If this is correct, and we want to go ahead with this, we might want to create shared utilities for doing this at some point. It's pretty heavy to set up as is.


demo/components/media-player/th90w160i1-samplevideo.png
svanherk, line 1: Looks like this file was missed in the migration, and there was a console error looking for it.

demo/components/wizard/index.html
```diff
@@ -1,15 +1,20 @@
...
-		import '@brightspace-ui/core/components/demo/demo-page.js';
+		import '../../../node_modules/@brightspace-ui/core/components/demo/demo-page.js';
 		import '../../../src/components/wizard/wizard.js';
 		import '../../../src/components/wizard/step.js';
 	</script>
+	<style>
+		d2l-labs-wizard {
+			line-height: 1.2rem;
+		}
+	</style>
```
svanherk, line 17: From what I can tell, the component doesn't actually work with our default line-height. I'm not going to touch the component right now, so fixing the demo.

rollup/rollup.config.js
```diff
@@ -0,0 +1,57 @@
...
+	'@(src|demo)/**/*.js',
+	'./index.js',
+	'node_modules/@brightspace-ui/core/components/calendar/*.js',
+	'node_modules/@brightspace-ui/core/components/collapsible-panel/*.js',
+	'node_modules/@brightspace-ui/core/components/colors/*.js',
+	'node_modules/@brightspace-ui/core/components/demo/*.js',
+	'node_modules/@brightspace-ui/core/components/dropdown/*.js',
+	'node_modules/@brightspace-ui/core/components/icons/*.js',
+	'node_modules/@brightspace-ui/core/components/menu/*.js',
+	'node_modules/@brightspace-ui/core/components/typography/*.js',
```
svanherk, line 22: Don't love this, but we need to tell rollup to include these components, which are _only_ used in the demos. This is likely only a labs repo problem.

src/components/accordion/README.md
svanherk, line 1: Didn't mean to include this as part of this PR lol, but 🤷

rollup/rollup.config.js
dlockhart: Wondering if you could use the `html` plugin to point the input at all the `.html` files [like this](https://github.com/dlockhart/brightspace-ui-core-demo/blob/main/rollup.config.js) and then `node-resolve` would properly resolve the bare module imports?

svanherk: That works! I knew I was missing something. It creates a bunch of `inline-module` js files, but I'm not concerned just for these previews.

dlockhart (approved):
Looking much better! Yeah I agree we probably want to start thinking about how to turn this into a shared action or something... it's a lot of duplication to wire up.

PR #637, opened 2025-05-14
GAUD-7905 - Rename slider-bar

svanherk:
This is only used by the media-player - just updating the name to match the rest of the repo.


dlockhart (approved):
It looks like we're not exporting this component, which is good I don't think we want to.

PR #644, opened 2025-05-21
GAUD-7908 - Add langterms from PR in original repo

svanherk:
These would eventually make it here, based on when they were added. But let's just move them now - copied from https://github.com/BrightspaceUILabs/grade-result/pull/407.


svanherk:
We'll want to include these in the backport alongside the other ones opened yesterday.

dlockhart (approved):

PR #645, opened 2025-05-21
GAUD-8024 - Set up minor releasing with the LMS

svanherk:
I noticed BSI has `2.24.1` and we've made lots of patch releases since then. Luckily they are all translation updates so it works out well, but I think we'd like this set up. It's been a while - will this "just work" and create the `json` file on the first run?


dlockhart (approved):
Haha not sure what it'll do!

svanherk:
It worked!

PR #646, opened 2025-05-21
GAUD-7947 - Cleanup grade-result docs

svanherk:
Cleaning up the README to create a nicer page for the Daylight site.

I moved a lot of the README contents into the files in case we ever decide to hook up `labs` to create the `custom-elements.json` files. But I can revert that part if we want and just leave it in the README only.

### DIFF
--- a/src/components/grade-result/grade-result-presentational.js
+++ b/src/components/grade-result/grade-result-presentational.js
@@ -14,30 +14,120 @@ const numberConverter = {
 	toAttribute:  (prop) => { return String(prop); }
 };
 
+/**
+ * @fires d2l-grade-result-grade-button-click - Dispatched when the grade button is clicked.
+ * @fires d2l-grade-result-manual-override-clear-click - Dispatched when the manual override clear is clicked.
+ * @fires d2l-grade-result-reports-button-click - Dispatched when the reports button is clicked.
+ */
 export class D2LGradeResultPresentational extends LocalizeLabsElement(LitElement) {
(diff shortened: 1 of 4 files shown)


src/components/grade-result/grade-result-presentational.js
```diff
@@ -14,30 +14,120 @@ const numberConverter = {
...
 			allowNegativeScore: { type: Boolean },
+			/**
+			 * This property will substitute the stock text on the "Clear Manual Override" button
+			 * @type {string}
+			 */
 			customManualOverrideClearText: { type: String },
+			/**
+			 * @type {boolean}
+			 */
 			displayStudentGradePreview: { type: Boolean, attribute: 'display-student-grade-preview' },
```
dlockhart, line 38: Pretty surprised this is the only one with `attribute`... eek.

dlockhart (approved):

src/components/grade-result/grade-result-presentational.js
svanherk: Yeah, we maybe should've updated them all during the migration, but that would've increased risk of a regression. If we still want to, it wouldn't be that much work. Most of the cleanup was removing it from repos _not_ using it.

dlockhart: Definitely riskier, and it looks like folks are using property binding in a few places to work around the weirdness. It'd be nice to fix if we think the number of consumers will grow...

PR #652, opened 2025-05-26
fix: backport translations to June release

dlockhart:
GAUD-8016

2/3 translation PRs already made it into `2.24.x` (but after branching), so this is just the one from today.


src/lang/haw.js
```diff
@@ -17,47 +17,47 @@ export default {
...
-	"components:mediaPlayer:seekSlider": "imi pahee",
-	"components:mediaPlayer:settings": "hoʻonohonoho",
-	"components:mediaPlayer:showSearchInput": "Hōʻike i ka hoʻokomo ʻimi",
-	"components:mediaPlayer:sliderBarProgress": "ka holomua ʻana o ka pahu slider",
-	"components:mediaPlayer:speed": "ka māmā holo",
+	"components:mediaPlayer:seekSlider": "seek slider",
+	"components:mediaPlayer:settings": "settings",
+	"components:mediaPlayer:showSearchInput": "Show search input",
+	"components:mediaPlayer:sliderBarProgress": "slider bar progress",
+	"components:mediaPlayer:speed": "speed",
```
svanherk, line 56: These are going back to English... 😕

dlockhart: What the... but just 5 of them...

dlockhart: Good catch -- I've reverted to the original values in `main` and applied that here as well.

PR #654, opened 2025-05-26
PHNX-2820 Ensure Accordion Initial State Accuracy

br-cz:
## Relevant Jira Links
https://desire2learn.atlassian.net/browse/PHNX-2820

## Description
The first time you navigate to the content page in Sequence Launcher,  the first accordion menu item is opened by default, but the `_state` attribute of the `d2l-labs-accordion-collapse` element is set to `closed`.

## Goal/Solution
The `_state` of the first item should be set to `opened`.

This is done by checking by checking if the item has been `opened`. In `connectedCallback` (which runs when the element is connected to the DOM), we set the `_state` to `'open'` if this is the case. This ensures the item's `_state` is set correctly.

## Screenshots/Demo
PHNX-2820

[PHNX-2820]: https://desire2learn.atlassian.net/browse/PHNX-2820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/accordion/accordion-collapse.js
+++ b/src/components/accordion/accordion-collapse.js
@@ -203,6 +203,9 @@ class LabsAccordionCollapse extends LitElement {
 			this.#resizeObserver = new ResizeObserver(() => this._fireAccordionResizeEvent());
 			this.#resizeObserver.observe(this);
 		}
+		if (this.opened) {
+			this._state = 'opened';
+		}
 	}
 	disconnectedCallback() {
 		super.disconnectedCallback();


src/components/accordion/accordion-collapse.js
```diff
@@ -203,6 +203,9 @@ class LabsAccordionCollapse extends LitElement {
 			this.#resizeObserver = new ResizeObserver(() => this._fireAccordionResizeEvent());
 			this.#resizeObserver.observe(this);
 		}
+		if (this.opened) {
+			this._state = 'opened';
+		}
```
br-cz, line 208: thought about using a ternary operator but since `_state` is already set to closed by the constructor, this looks better in my opinion but definitely open to changing!

dlockhart: `_state` seems to be controlled completely by the user interacting with the `<d2l-expand-collapse-content>`, i.e. it's not currently impacted by `opened` at all. So I think things can still get out of sync if a consumer changes the value of `opened` after this initial `connectedCallback` runs? Seems like it might be delicate to fix that though as there's a lot going on with the various events and such.

br-cz: I've given it some thought and to me it seems like the `<d2l-expand-collapse-content>` component is triggered alongside any changes to `opened`, which would in turn change `_state` to an appropriate value. So even though a consumer may change the value of `opened` after the initial callback run, it will change to the appropriate value. But, of course I could be misinterpreting the code and have a total misunderstanding - as you mentioned there is a lot going with the various event 😓. 

I'd love to hear your thoughts, and excited to attend the D2T class tomorrow! 🥳

PR #671, opened 2025-06-09
GAUD-8069: updated router demo/docs for Daylight Site

dlockhart:


demo/utilities/router/app.js
```diff
@@ -29,9 +27,9 @@ export class App extends LitElement {
 			<aside>
 				<nav>
 					<ul>
-						<li><a href="/example/">Home</a></li>
-						<li><a href="/example/people">People</a></li>
-						<li><a href="/example/places">Places</a></li>
```
dlockhart, line 34: The old repo had the demo in `/example`, which doesn't work anymore.

demo/utilities/router/data.js
```diff
@@ -0,0 +1,77 @@
+export const people = [
```
dlockhart, line 1: Split the data out into a separate file.

demo/utilities/router/route-loader.js
```diff
@@ -21,6 +17,13 @@ registerRoutes(
...
 		placesRouteLoader,
+		{
+			pattern: '*',
+			view: () => html`<h1>Not Found</h1>`,
+		}
 	],
-	{ basePath: '/example' }
+	{
+		basePath: '/demo/utilities/router',
+		enableRouteOrderFix: true
```
dlockhart, line 27: It's unfortunate that new uses always need to enable this. I'm tempted to create a flag that does the inverse (disables the fix), switch all existing places to set that, and then change the behaviour so the fix is applied by default.

svanherk: That would make sense to me

svanherk:
The live link to the router demo goes somewhere a bit wonky to start, but then is good once you pick a page: https://live.d2l.dev/prs/BrightspaceUI/labs/pr-671/demo/utilities/router/

src/utilities/router/README.md
```diff
@@ -59,20 +48,20 @@ The view is provided a context object containing:
...
   view: ctx => html`
-    <user-view
-        id=${ctx.options.id}
-        page=${ctx.params.page}
-        semester=${ctx.search.semester}>
-    </user-view>
+    <my-app-user
+        id="${ctx.options.id}"
+        page="${ctx.params.page}"
+        semester="${ctx.search.semester}>"
```
svanherk, line 56:
```suggestion
        semester="${ctx.search.semester}">
```

svanherk (approved):

svanherk (approved):
Awesome!

PR #672, opened 2025-06-09
GAUD-8069: add flag to disable route order fix

dlockhart:
We'd prefer that the route order fix be applied by default so that consumers don't need to worry about it, we don't need to warn them about it and we can generally move on with our lives. 😆 

I'm going to do this in 4 phases:
1. Introduce `disableRouteOrderFix`. When set to `true`, the fix won't be applied.
2. Update consumers that aren't currently setting `enableRouteOrderFix` to pass `disableRouteOrderFix`.
3. Remove the concept of a `enableRouteOrderFix` option, so the fix will be enabled by default.
4. Update consumers that were sending `enableRouteOrderFix` to no longer pass it.

### DIFF
--- a/test/utilities/router/router.test.js
+++ b/test/utilities/router/router.test.js
@@ -55,7 +55,11 @@ const initRouter = () => {
 			load1,
 			load2,
 		],
-		{ hashbang: true, customPage: true }
+		{
+			enableRouteOrderFix: true,
+			hashbang: true,
+			customPage: true
+		}
 	);
 };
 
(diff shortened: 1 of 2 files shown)


test/utilities/router/router.test.js
```diff
@@ -55,7 +55,11 @@ const initRouter = () => {
 			load1,
 			load2,
 		],
-		{ hashbang: true, customPage: true }
+		{
+			enableRouteOrderFix: true,
```
svanherk, line 59: Does adding this here do anything if there's no wildcards?

svanherk (approved):

PR #678, opened 2025-06-12
GAUD-8027: switch to kebab-case attributes

dlockhart:
Many of the `<d2l-labs-grade-result-*>` components don't adhere to our kebab-case guidance for multi-word attributes, and we want to resolve that.

This change is going to happen in 3 stages:
1. Go through consumers of the components and add both kebab-case and non-kebab-case attributes (complete ✅)
2. Switch the `<d2l-labs-grade-result-*>` components to only support kebab-case (this PR)
3. Go back through consumers and remove the non-kebab-case attributes

### DIFF
--- a/test/components/grade-result/d2l-grade-result-student-grade-preview.vdiff.js
+++ b/test/components/grade-result/d2l-grade-result-student-grade-preview.vdiff.js
@@ -1,6 +1,7 @@
 import '../../../src/components/grade-result/grade-result-student-grade-preview.js';
 import '../../../src/components/grade-result/grade-result-presentational.js';
 import { expect, fixture, html } from '@brightspace-ui/testing';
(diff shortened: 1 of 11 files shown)


test/components/grade-result/d2l-grade-result-student-grade-preview.vdiff.js
```diff
@@ -176,7 +177,6 @@ describe('presentational-with-grade-preview', () => {
 			labelText: 'Overall Grade',
 			scoreNumerator: '5',
 			scoreDenominator: '20',
-			isGradeAutoCompleted: true,
```
dlockhart, line 179: This [attribute was removed](https://github.com/BrightspaceGraveyard/grade-result/pull/85) so does nothing.

svanherk (approved):

PR #729, opened 2025-08-12
NTNGL-4224 | Loader can access Context

NicholasHallman:
This allows us to use the loader as a means of loading other kinds of data other than just js modules. e.g.

``` javascript
registerRoutes([
	{
		pattern: '/d2l/ap/InsightsPortal/:orgUnitId/:tab',
		loader: async(ctx) => {
			const importPromise = import('./app.js');
			const dataPromise = (async() => {
				const result = await fetch('/need/more/data');
				ctx.loaderData= await result.json();
			})();
			await Promise.all([dataPromise, importPromise]);
		},
		view: ({ loaderData}) => html`<d2l-my-app .loaderData=${options.loaderData}></d2l-my-app>`
	}
], {});
```

### DIFF
--- a/demo/utilities/router/route-loader.js
+++ b/demo/utilities/router/route-loader.js
@@ -8,8 +8,11 @@ registerRoutes(
 	[
 		{
 			pattern: '/',
-			loader: () => import('./home.js'),
-			view: () => html`<test-home></test-home>`,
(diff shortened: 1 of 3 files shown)


dlockhart (approved):
Neat!

PR #767, opened 2025-09-16
US15489 | Community Link uses Core Link

NicholasHallman:
https://desire2learn.atlassian.net/browse/NTNGL-4891

We were using a clone of the link component while waiting for new window icon support on the core link. I forgot to go back and replace the temp link with the real one though!


svanherk (approved):
Yay! 🎉

PR #795, opened 2025-10-17
Update `@web/dev-server`

devpow112:
Should get rid of a [vulnerability](https://github.com/BrightspaceUI/labs/security/dependabot/8). Was an easy update as nothing we used changed. The rest of the package changes just removed the misleading bits. We are only getting the latest of the majors (`^`) for the various packages changed.

https://desire2learn.atlassian.net/browse/QE-1260


dlockhart (approved):
Amazing, thanks!

PR #800, opened 2025-10-21
GAUD-8826: revert page.js fork

dlockhart:
Reverting #797, as projects that have their own copy of `page.js` (like Folio) end up in a conflicted state and don't have access to the `window` object. I think it's [related to this issue](https://github.com/visionmedia/page.js/issues/537).

We may need to fork `page.js` afterall and publish our own patched version of it and switch everything over to that.


eKoopmans (approved):
Is `fix: ` prefix needed?

svanherk (approved):
I'll follow up in slack with more questions. Does this need a backport?

dlockhart:
> Does this need a backport?

It merged Friday afternoon so I'm hoping not -- not for Folio anyway right @eKoopmans?

PR #877, opened 2026-01-19
Add pre and post navigation hooks to router

duncanuszkay-d2l:
[Ticket](https://desire2learn.atlassian.net/browse/NTNGL-5196?atlOrigin=eyJpIjoiMzEzMGU0YmQ2ZmEyNDY1Zjg3YWVkYjNjOTgxNDU2ZGEiLCJwIjoiaiJ9)

Following up on [Nick's work](https://github.com/BrightspaceUI/labs/pull/842), adding pre and post navigation hooks that execute before and after the page routing executes based on @dlockhart 's commentary.

The feature is described in greater detail in the `README`

Example PR utilizing this new feature downstream: https://github.com/Brightspace/insights/pull/1265/files

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@brightspace-ui/labs",
-  "version": "2.43.0",
+  "version": "2.43.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@brightspace-ui/labs",
-      "version": "2.43.0",
+      "version": "2.43.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/lit-mobx": "^2",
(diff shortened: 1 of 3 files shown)


src/utilities/router/router.js
```diff
@@ -85,6 +85,37 @@ const _registerRoutes = routes => {
...
+		next();
+	});
+};
+
+const _registerPostNavigateHooks = () => {
+	activePage('*', (context, next) => {
+		RouterHooks.postNavigate.forEach(h => h(context));
+		next();
+	});
+};
```
duncanuszkay-d2l: This could be folded in by keying on a string hook name instead of having separate methods, but I find it cleaner to keep them discretely listed as functions. No strong opinion.

NicholasHallman:
Nice work, very straightforward. It would be nice if there was a test to check if the pre and post hooks are getting called in the right order

test/utilities/router/router.test.js
```diff
@@ -80,7 +95,7 @@ describe('Router', () => {
 		entryPoint = await fixture(
 			html`<main-view main-prop="Passed"></main-view>`
 		);
-		navigate('/');
+		// navigate('/');
```
dlockhart: Probably didn't intend to leave this commented out?

NicholasHallman (approved):

PR #969, opened 2026-04-23
GAUD-9791: fix aria-controls

dlockhart:
The `aria-controls` here is pointing at nothing, which was being caught in Lighthouse and axe. Looks like it was broken [about 6 years ago!](https://github.com/BrightspaceUI/labs/commit/6af46fdf754837cef6be860084ef51fb8ad73e2d)! 😬 

Also fixing `aria-expanded`, which is a string not a boolean.

### DIFF
--- a/src/components/accordion/accordion-collapse.js
+++ b/src/components/accordion/accordion-collapse.js
@@ -221,7 +221,7 @@ class LabsAccordionCollapse extends LitElement {
 	render() {
 		return html`
 			<div id="header-container">
-				<a href="javascript:void(0)" id="trigger" ?aria-expanded=${this.opened} class="header-grid-item" aria-controls="collapse" role="button" ?data-border="${this.headerBorder}" @blur=${this._triggerBlur} @click=${this.toggle} @focus=${this._triggerFocus}>
+				<a href="javascript:void(0)" id="trigger" aria-expanded=${this.opened ? 'true' : 'false'} class="header-grid-item" aria-controls="collapse" role="button" ?data-border="${this.headerBorder}" @blur=${this._triggerBlur} @click=${this.toggle} @focus=${this._triggerFocus}>
 					${!this.headerHasInteractiveContent ? html`
 						<div class="collapse-title" title="${this.label}">${this.title}${this.label}<slot name="header"></slot>
 						</div>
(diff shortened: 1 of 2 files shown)


margaree (approved):

src/components/accordion/accordion-collapse.js
```diff
@@ -243,7 +243,7 @@ class LabsAccordionCollapse extends LitElement {
 				` : nothing}
 			</div>
 
-			<div class="content">
+			<div class="content" id="collapse">
```
dbatiste, line 246: Should the `aria-controls` point to this or the thing that's actually expanded/collapsed?

dlockhart: Well that's nested deep in `d2l-expand-collapse-content`'s shadow DOM 😆 

I can move it to `<d2l-expand-collapse-content>`, but it doesn't make a difference since it's also not the thing that hides/shows.

dbatiste: Ah I see.

dbatiste (approved):

src/components/accordion/accordion-collapse.js
```diff
@@ -221,7 +221,7 @@ class LabsAccordionCollapse extends LitElement {
 	render() {
 		return html`
 			<div id="header-container">
-				<a href="javascript:void(0)" id="trigger" ?aria-expanded=${this.opened} class="header-grid-item" aria-controls="collapse" role="button" ?data-border="${this.headerBorder}" @blur=${this._triggerBlur} @click=${this.toggle} @focus=${this._triggerFocus}>
+				<a href="javascript:void(0)" id="trigger" aria-expanded=${this.opened ? 'true' : 'false'} class="header-grid-item" aria-controls="collapse" role="button" ?data-border="${this.headerBorder}" @blur=${this._triggerBlur} @click=${this.toggle} @focus=${this._triggerFocus}>
```
bearfriend, line 224: Not worth changing, but fyi
```suggestion
				<a href="javascript:void(0)" id="trigger" aria-expanded=${String(this.opened)} class="header-grid-item" aria-controls="collapse" role="button" ?data-border="${this.headerBorder}" @blur=${this._triggerBlur} @click=${this.toggle} @focus=${this._triggerFocus}>
```

PR #983, opened 2026-05-08
GAUD-9877: AI Lumi button

EdwinACL831:
## Jira
[GAUD-9877](https://desire2learn.atlassian.net/browse/GAUD-9877)

## Description
This addresses the intention of creating a specific AI Lumi Button. This is done by using the `d2l-button-subtle` from core plus adding styles to it.

NOTE: Still a WIP

### Normal state
image

### Focused state (by tabbing from keyboard)
image

### Focused state (by focusing programmatically and by clicking it)
`Focused by clicking another button that focuses the AI one` also by `clicking the ai button directly`

image

### Hover state (before any click)
image

### Disabled (no tooltip)
image

### Disabled (with tooltip + focus)
image

[GAUD-9877]: https://desire2learn.atlassian.net/browse/GAUD-9877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


demo/components/button-subtle-ai/index.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
+		<script type="module">
+			import '@brightspace-ui/core/components/demo/demo-page.js';
+			import "../../../src/components/button-subtle-ai/button-subtle-ai.js";
+		</script>
+		<title>d2l-labs-button-subtle-ai</title>
```
dbatiste: This `title` element is redundant since `d2l-demo-page` creates a title element using the `page-title` provided.

src/components/button-ai/button-ai.js
dlockhart, line 1: Question for Gaudi: wondering whether we might want an `ai` directory where we can put all the AI-related stuff? Not sure if there are plans for more components or not. Otherwise we'd end up with `button-ai`, `dropdown-ai`, etc.

dbatiste, line 1: I would be tempted to put this in a `\button-ai` folder instead of `\button-subtle-ai`, so that if they eventually add any other button variant, we can keep them grouped.

src/components/button-subtle-ai/button-subtle-ai.js
```diff
@@ -0,0 +1,40 @@
...
+			 * ACCESSIBILITY: REQUIRED: Text for the button
+			 * @type {string}
+			 */
+			text: { type: String, reflect: true }
+		};
+	}
+
+	static get styles() {
+		return [css`
+			:host d2l-button-subtle {
```
dlockhart: I don't think we want to be styling core components like this, as it's going to be super fragile when their styles change. Also a lot of the styles need to apply to `d2l-button-subtle`'s internal `<button>` element.

It might be easier actually to just _not_ use `d2l-button-subtle` and render our own `<button>` here...

test/components/button-subtle-ai/button-subtle-ai.test.js
```diff
@@ -0,0 +1,24 @@
...
+import { html } from 'lit';
+
+describe('button-subtle-ai', () => {
+	describe('constructor', () => {
+		it('should construct', () => {
+			runConstructor('d2l-labs-button-subtle-ai');
+		});
+	});
+
+	describe('rendering', () => {
```
dlockhart: I'd probably remove these tests as they're not really testing anything useful. If the text/icon isn't set properly then the vdiffs will fail.

test/components/button-subtle-ai/button-subtle-ai.vdiff.js
```diff
@@ -0,0 +1,13 @@
+import '../../../src/components/button-subtle-ai/button-subtle-ai.js';
+import { expect, fixture, focusElem } from '@brightspace-ui/testing';
+import { html } from 'lit';
+
+describe('button-subtle-ai', () => {
+	describe('states', () => {
+		it('focus', async() => {
+			const element = await fixture(html`<d2l-labs-button-subtle-ai text="Test"></d2l-labs-button-subtle-ai>`, { viewport: { width: 150, height: 100 } });
```
dlockhart: I don't think you need to set the viewport here if you pass `element` to `expect()`.

src/components/button-subtle-ai/button-subtle-ai.js
```diff
@@ -0,0 +1,40 @@
...
+				border: 2px solid transparent;
+				border-radius: 6px;
+				background:
+					linear-gradient(var(--d2l-theme-background-color-base), var(--d2l-theme-background-color-base)) padding-box,
+					linear-gradient(to top left, var(--d2l-color-celestine), var(--d2l-color-fluorite-plus-1)) border-box;
+
+				&:focus-within {
```
dbatiste: Are we able to use nesting selectors in CSS? [canIuse](https://caniuse.com/css-nesting).

test/components/button-subtle-ai/button-subtle-ai.vdiff.js
```diff
@@ -0,0 +1,13 @@
+import '../../../src/components/button-subtle-ai/button-subtle-ai.js';
+import { expect, fixture, focusElem } from '@brightspace-ui/testing';
+import { html } from 'lit';
+
+describe('button-subtle-ai', () => {
+	describe('states', () => {
+		it('focus', async() => {
```
dbatiste: I think we would want to capture other states as well, since the special styling is only really noticeable in the normal and hover states.

test/components/button-subtle-ai/button-subtle-ai.vdiff.js
dbatiste: I think we can avoid the special viewport setting here and we can just capture the element.
```javascript
await expect(element).to.be.golden();
```

src/components/button-subtle-ai/button-subtle-ai.js
```diff
@@ -0,0 +1,40 @@
+
+import '@brightspace-ui/core/components/button/button-subtle.js';
+import { css, html, LitElement } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+class ButtonSubtleAI extends LitElement {
```
dbatiste: I think this should extend the `FocusMixin` so that anything that calls focus on it will apply `focus` to the underlying element.

```javascript
static get focusElementSelector() {
	return 'd2l-button-subtle';
}
```

test/components/button-subtle-ai/button-subtle-ai.vdiff.js
```diff
@@ -0,0 +1,13 @@
...
+			const element = await fixture(html`<d2l-labs-button-subtle-ai text="Test"></d2l-labs-button-subtle-ai>`, { viewport: { width: 150, height: 100 } });
+			await focusElem(element.shadowRoot.querySelector('d2l-button-subtle'));
```
dbatiste: Once the `focus` method is implemented (via `FocusMixin`), I think we should be able to just use:
```javascript
await focusElem(element);
```

test/components/button-subtle-ai/button-subtle-ai.test.js
dbatiste: These seem redundant. I think the visual-diff captures setting the `text` and the `icon` already.

test/components/button-subtle-ai/button-subtle-ai.vdiff.js
EdwinACL831: This is how it looks like by passing element to the expect(), kind of chopped:

Image

dlockhart: Try making the `:host` `display: inline-block;`.

EdwinACL831: Think I found the issue, is because the actual `d2l-labs-button-subtle-ai` component doesn't have the same dimension as the `d2l-button-subtle`

Update: did this and worked
```javascript
const element = await fixture(html`<d2l-labs-button-subtle-ai text="Test"></d2l-labs-button-subtle-ai>`);
const button = element.shadowRoot.querySelector('d2l-button-subtle');
await hoverElem(button);
await expect(button).to.be.golden();
```

will try what you suggest

dlockhart: Yep -- custom elements are `display: inline` by default which means they have no height. Setting it to `display: inline-block` will fix it.

src/components/button-subtle-ai/button-subtle-ai.js
EdwinACL831: Will double check this.

src/components/button-ai/button-ai.js
EdwinACL831: I think that makes sense, is very likely that more requests like this one to come up

src/components/button-subtle-ai/button-subtle-ai.js
EdwinACL831: this has been changed, please feel free to revisit the code

test/components/button-subtle-ai/button-subtle-ai.vdiff.js
EdwinACL831: added some more

src/components/button-ai/button-ai.js
bearfriend: I think a `lumi` directory makes sense, and I'm not sure how "subtle" this is anymore

```shell
./src/components/lumi/button/button.js
```

```javascript
class Button extends FocusMixin(LitElement) {
...
customElements.define('d2l-labs-lumi-button', Button);
```

```javascript
import '@brightspace-ui/labs/components/lumi/button/button.js';
```
```html
<d2l-labs-lumi-button text="Rot My Brain"></d2l-labs-lumi-button>
```

src/components/button-ai/button-subtle-ai.js
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	static get styles() {
+		return [css`
+			:host {
+				background:
+					linear-gradient(var(--d2l-theme-background-color-base), var(--d2l-theme-background-color-base)) padding-box,
+					linear-gradient(to top left, var(--d2l-color-celestine), var(--d2l-color-fluorite-plus-1)) border-box;
+				border: 0.1rem solid transparent;
+				border-radius: 0.3rem;
```
bearfriend: Image
<br>
The radius needs to increase with the border width to match the inner button shape/focus styles

```suggestion
				border-radius: 0.4rem;
```

bearfriend:
Are we ok with this button's box being larger than our standard buttons? I don't see a way to shrink it without extending `ButtonMixin` directly

Screenshot 2026-05-08 at 9 56 52 PM

src/components/button-ai/button-ai.js
```diff
@@ -0,0 +1,45 @@
...
+			 * ACCESSIBILITY: REQUIRED: Text for the button
+			 * @type {string}
+			 */
+			text: { type: String, reflect: true }
+		};
+	}
+
+	static get styles() {
+		return [css`
+			:host {
```
bearfriend, line 21: We probably want a wrapper in the shadow DOM to style so they're scoped and not easily overridden by consumers

src/components/button-ai/button-ai.js
EdwinACL831: IMO that sounds like very product specific, and not sure if this repo is intended for that, like if we would have a folder or even a label specific/dedicated to a product. To me sounds like we should not, but we could talk about this within the team's channel.

I think @dbatiste suggested something similar but more generic: `button-ai` which to me sounds like the way to go

bearfriend: Yeah definitely not saying you're wrong. I guess my thinking with using `lumi` was to discourage wider usage. This is the branding of Lumi, right? Maybe I just need to better understand where the lines are.

Does Lumi have its own defined design pattern as a branch off Daylight?

Are there/will there be other AI tools/products that aren't Lumi-branded? Will they all use this design pattern?

Will this graduate to core or somewhere else? If somewhere else, to a product repo like nearly everything else is organized?

geurts:
We met with the Lumi X designers, and they have no plans to use this approach for other components (they dropped their plans to do so with d2l-dropdown since it is not feasible without refactoring the core component and they don't have sufficient design explorations to support a pattern). So we have agreed to continue with the plan of creating this as a one-off Labs component for their use. If a pattern emerges when folks have time to properly think this through, we'll consider other ways to support AI indicators in Core more systematically.

@EdwinACL831 @nathankoop — I think we can proceed with wrapping this up and handing it off to Lumi X devs, unless I'm missing something?

src/components/button-ai/button-ai.js
```diff
@@ -0,0 +1,96 @@
...
+				padding-block: 0;
+				padding-inline-end: var(--d2l-button-subtle-padding-inline-end);
+				padding-inline-start: var(--d2l-button-subtle-padding-inline-start);
+				position: relative;
+			}
+			button:hover,
+			button:focus,
+			:host([active]) button {
+				background-color: var(--d2l-theme-background-color-interactive-tertiary-hover);
+			}
```
bearfriend, line 53: Something like these `min-height` values should keep it the same size as our other buttons

```suggestion
				font-family: inherit;
				min-height: calc(2px + 2rem - .2rem);
				padding-block: 0;
				padding-inline-end: var(--d2l-button-subtle-padding-inline-end);
				padding-inline-start: var(--d2l-button-subtle-padding-inline-start);
				position: relative;
			}
			button:hover,
			button:focus,
			:host([active]) button {
				background-color: var(--d2l-theme-background-color-interactive-tertiary-hover);
				min-height: calc(2px + 2rem);
			}
```

src/components/button-ai/button-ai.js
bearfriend: ^ Should still do this

dlockhart: Agreed -- I think we just want to apply these to the `<button>` element and have nothing on the host?

src/components/button-ai/button-subtle-ai.js
```diff
@@ -0,0 +1,96 @@
...
+			:host([active]:not([disabled])) slot[name="icon"]::slotted(d2l-icon-custom) {
+				color: var(--d2l-theme-text-color-interactive-hover);
+			}
+	`];
+
+	constructor() {
+		super();
+		this.icon = 'tier1:ai';
+		this.disabledTooltip = 'This feature is not yet available.';
+		this._buttonId = getUniqueId();
```
dlockhart: Could be a private `#buttonId` variable.

src/components/button-ai/button-subtle-ai.js
```diff
@@ -0,0 +1,96 @@
...
+			<button ?disabled=${this.disabled} id="${this._buttonId}" class="d2l-label-text">
+				${this._renderIcon()}
+				<span class="d2l-button-subtle-content">${this.text}</span>
+			</button>
+			${this.disabled && this.disabledTooltip ? html`<d2l-tooltip class="vdiff-target" for="${this._buttonId}">${this.disabledTooltip}</d2l-tooltip>` : ''}
+		`;
+	};
+};
+
+customElements.define('d2l-labs-button-subtle-ai', ButtonSubtleAI);
```
dlockhart: Now that this isn't using/extending subtle button, should it just be `<d2l-labs-button-ai>`?

src/components/button-ai/button-subtle-ai.js
```diff
@@ -0,0 +1,96 @@
...
+			button:hover:not([disabled]) slot[name="icon"]::slotted(d2l-icon-custom),
+			button:focus:not([disabled]) slot[name="icon"]::slotted(d2l-icon-custom),
+			:host([active]:not([disabled])) slot[name="icon"]::slotted(d2l-icon-custom) {
+				color: var(--d2l-theme-text-color-interactive-hover);
+			}
+	`];
+
+	constructor() {
+		super();
+		this.icon = 'tier1:ai';
```
dlockhart, line 91: Technically this will let consumers set the icon to whatever they want. Maybe not a big deal, but worth noting.

src/components/button-ai/button-subtle-ai.js
```diff
@@ -0,0 +1,96 @@
...
+		this.disabledTooltip = 'This feature is not yet available.';
```
dlockhart: Is it not? Looks like it's supported.

src/components/button-ai/button-ai.js
EdwinACL831: I think the min-height is being applied by `buttonStyles` in [this line](https://github.com/BrightspaceUI/labs/pull/983/changes#diff-5e7231e0cd77cf5497bde688e744c76033bc7a7ceef771922fcff2657389e839R18). So it comes from there and merges with these styles here

src/components/button-ai/button-subtle-ai.js
EdwinACL831: Oh yeah, I was doing some testing here 😅 I wasn't sure what should be the disabled tooltip message. think will just remove it

src/components/button-ai/button-ai.js
bearfriend: Right, it needs to be overridden

src/components/button-ai/button-subtle-ai.js
EdwinACL831: So the way i thought on this was to have this as the default and they can use their own custom

src/components/button-ai/button-ai.js
EdwinACL831: Sorry but it look wonky, are they suppose to have 2 different min-height values ? or do you mean to override the deafult with the `min-height: calc(2px + 2rem - 0.2rem);` on each ?

EdwinACL831: Also, I think that @geurts mentioned to always have a 42px height which is the one by default

src/components/button-ai/button-ai.js
EdwinACL831: I have added a wrapper for the button that holds those styles now

src/components/button-ai/button-subtle-ai.js
EdwinACL831: removed

src/components/button-ai/button-ai.js
bearfriend: Sorry I think we're leading you in circles a bit. The button currently is _not_ 42px because of the added border. Overriding the `min-height` to subtract  the height of the border will fix that. The second one for the focus styles I think isn't necessary with the newer design since it keeps the border on focus.

Additionally, in @dlockhart's comment above, since you are now extending `ButtonMixin` directly, he suggested applying all the `:host` (now `.d2l-labs-button-ai-container`) styles to the `button` itself, which may actually remove the need for this override anyway.

Lots of moving parts here. Let me know if you need more clarification.

Screenshot 2026-05-27 at 2 04 11 PM


**^ Edited this comment significantly**

bearfriend: Confirmed, if you remove the container and apply all the styles to the `button`, then you don't need to override `min-height` _or_ `--d2l-focus-ring-offset`

EdwinACL831: @bearfriend thanks for the very detailed explanation here. I see the problem and actually from my end - in the browser - I was looking at the internal button and not the actual d2l-button-ai component. Therefore of course the button was 42px tall but - again, as you correctly explained - the whole web component was 46px in total.

I did the update of the code and yes, the min-height override is not needed after that.
Also not sure but I think the `ButtonMixin` only handles some behavior of the button (like the click event and the focus-visible event too. The styles come from the `buttonStyles` which allows me to also do the proposed solution.

src/components/button-ai/button-ai.js
```diff
@@ -0,0 +1,101 @@
...
+					linear-gradient(to top left, var(--d2l-color-fluorite-plus-1), var(--d2l-color-celestine-plus-1)) border-box;
+			}
+
+			button:hover:not([disabled]) .d2l-button-subtle-content,
+			button:focus:not([disabled]) .d2l-button-subtle-content,
+			:host([active]:not([disabled])) button .d2l-button-subtle-content {
+				color: var(--d2l-theme-text-color-interactive-hover);
+			}
+
+			button:hover:not([disabled]),
```
EdwinACL831: Not sure, but it looks like these CSS rules are not really necessary. @bearfriend and @dlockhart I will remove them but before I will let you guys give your point of view on this, if you are willing to.

dlockhart: So I think they might be needed, but the reason they don't appear needed is I don't think the `disabled` attribute is working properly right now. Like when I set it, none of the styles seem to change.

I think what you'll want to do is set an `opacity` (like subtle button does) when the button is disabled, and then also set the background back to the default when one it's disabled (also like subtle button does).

dlockhart: And I think you'll want to add vdiffs for when it's disabled, including at least hovering over a disabled button.

src/components/button-ai/button-ai.js
bearfriend: Looks good!

src/components/button-ai/button-ai.js
```diff
@@ -0,0 +1,86 @@
...
+				font-family: inherit;
+				padding-block: 0;
+				padding-inline: 0.6rem;
+				position: relative;
+			}
+
+			button:hover,
+			button:focus {
+				background:
+					linear-gradient(var(--d2l-theme-background-color-interactive-tertiary-hover), var(--d2l-theme-background-color-interactive-tertiary-hover)) padding-box,
```
dbatiste: I think we'll want to be careful here. These variables aren't "locked in", and there is a good chance they will change, requiring follow-up PRs so this does not break.

src/components/button-ai/button-ai.js
EdwinACL831: I tried so I can also make sure the tooltip displays if there is any. It was not being captured on the ss of the vdiff. I went to the doc and found that I had to add the `vdiff-include` class to the tooltip. Did it but did not worked. also tried the `vdiff-target` but same unsuccessful result.

src/components/button-ai/button-ai.js
EdwinACL831: OK I will look into which ones we would like to have as css vars and which ones we would like to actually have their values. I think most of them would fall into the second case.

EdwinACL831: I think these variables since are part of the color palette in order to support dark mode, they should be used. Otherwise when enabling dark mode in the LMS, this button would have fixed values that may not fall under AX criteria and the dark mode designs criteria imho

dlockhart: I think Dave B's question is more around whether we support dark mode out of the box but risk having it break if we rename the variables (since they're still changing), or if we don't support it initially but have to come back and add support later. So far `@brightspace-ui/core` is the only repo we've been adding dark mode support to.

src/components/button-ai/button-ai.js
```diff
@@ -46,6 +47,30 @@ class ButtonAI extends SlottedIconMixin(ButtonMixin(LitElement)) {
 					linear-gradient(to top left, var(--d2l-color-fluorite-plus-1), var(--d2l-color-celestine-plus-1)) border-box;
 			}
 
+			:host([disabled]) button {
+				cursor: default;
+			}
+
+
+			:host([disabled]) button:hover,
+			:host([disabled]) button:focus {
```
dlockhart: Since this is a duplication of the `background` for `button`, can we combine all 3 selectors?

```css
button,
:host([disabled]) button:hover,
:host([disabled]) button:focus {
	background:
		linear-gradient(var(--d2l-theme-background-color-base), var(--d2l-theme-background-color-base)) padding-box,
		linear-gradient(to top left, var(--d2l-color-fluorite-plus-1), var(--d2l-color-celestine-plus-1)) border-box;
}
```

src/components/button-ai/button-ai.js
```diff
@@ -46,6 +47,30 @@ class ButtonAI extends SlottedIconMixin(ButtonMixin(LitElement)) {
...
+			:host([disabled]) button::before {
+				background:
+					linear-gradient(var(--d2l-theme-background-color-base), var(--d2l-theme-background-color-base)) padding-box,
+					linear-gradient(to top left, var(--d2l-color-fluorite-plus-1), var(--d2l-color-celestine-plus-1)) border-box;
+				border-radius: inherit;
+				content: '';
+				inset: 0;
+				opacity: 0.5;
+				position: absolute;
+			},
```
dlockhart: Can we replace this with something like this? This is what subtle button does.

```css
:host([disabled]) button {
		cursor: default;
		opacity: var(--d2l-theme-opacity-disabled-control);
}
```

src/components/button-ai/button-ai.js
```diff
@@ -46,6 +47,30 @@ class ButtonAI extends SlottedIconMixin(ButtonMixin(LitElement)) {
...
+
 			button:hover:not([disabled]) .content,
 			button:focus:not([disabled]) .content {
```
dlockhart: These rules that use `:not([disabled])` on the `button` don't work when the `disabled-tooltip` is used, because in that scenario the underlying button never gets the `disabled` attribute -- it gets `aria-disabled` instead.

This should work though:

```css
:host(:not([disabled])) button:hover .content
```

demo/components/button-ai/index.html
```diff
@@ -42,6 +42,16 @@ <h2>Button AI (custom image with icon attribute)</h2>
...
 			</d2l-demo-snippet>
+
+			<h2>Button AI (disabled)</h2>
+			<d2l-demo-snippet>
+				<d2l-labs-button-ai text="New AI Feature" disabled></d2l-labs-button-ai>
+			</d2l-demo-snippet>
+
+			<h2>Button AI (disabled with tooltip)</h2>
+			<d2l-demo-snippet>
+				<d2l-labs-button-ai text="New AI Feature" disabled disabled-tooltip="No available"></d2l-labs-button-ai>
```
dlockhart: Should be `"Not available"`

test/components/button-ai/button-ai.vdiff.js
```diff
@@ -28,5 +28,10 @@ describe('button-ai', () => {
 			await clickElem(element);
 			await expect(element).to.be.golden();
 		});
+
+		it('disabled', async() => {
+			element.disabled = true;
+			await expect(element).to.be.golden();
+		});
```
dlockhart, line 44: Can we have tests for disabled + hover and focus, as well as disabled + disabled-tooltip while hovered and focused? I think it would have caught the bug here where hovering over a disabled-tooltip button is broken currently.

src/components/button-ai/button-ai.js
EdwinACL831: hmm ok then. I think we should go with the second option: Making the button to work for regular light mode and then add the dark mode support when it is stable I guess.

src/components/button-ai/button-ai.js
EdwinACL831: I think this was a comment previous my last update. I actually added this before getting into this comment.
This is to style the button when the component is disabled but the button is still focusable (e.g., when having a tooltip). In that case I added only the cursor: default and the opacity I moved that to this selector: `:host([disabled]) button::before` because that way I was able to make the opacity goes only in the button and not the focus ring (as per design requested)

src/components/button-ai/button-ai.js
EdwinACL831: Yeah I fall into this and figured that out. These changes should reflect what you are suggesting

test/components/button-ai/button-ai.vdiff.js
EdwinACL831: just not understanding why the tooltip is no added to the ss even using the vdiff-include class. and also when hovering over the disabled button with tooltip, it doesn't appear either

src/components/button-ai/button-ai.js
EdwinACL831: OK so I have used only the regular color pallet which is stable, and not the semantic one which it isn't yet.

package.json
```diff
@@ -8,6 +8,7 @@
     "./components/accordion.js": "./src/components/accordion/accordion.js",
     "./components/accordion-collapse.js": "./src/components/accordion/accordion-collapse.js",
     "./components/attribute-picker.js": "./src/components/attribute-picker/attribute-picker.js",
+    "./components/button-subtle-ai.js": "./src/components/button-ai/button-subtle-ai.js",
```
dlockhart:
```suggestion
    "./components/button-ai.js": "./src/components/button-ai/button-ai.js",
```

EdwinACL831: nice catch, thanks

dlockhart (approved):
LGTM!

PR #990, opened 2026-05-12
MED-2833: Suppress benign AbortError

rcausey:
Suppress an intermittent AbortError in the media-player to reduce console noise. The error is difficult to reproduce. I was able to reproduce it a few times with the following steps:

1. Pressing Play
2. Pressing Pause
3. Waiting for the media URL to expire
4. Pressing Play

In this scenario, playback recovered successfully by loading a new source with no user impact. 

Subsequent attempts to reproduce the error - including simulating an immediate pause after play - were unsuccessful, and it's unclear if other scenarios exist that can trigger it.

With no known user impact and no related support tickets, this appears to be a benign, self-recovering error and is safe to suppress.

### DIFF
--- a/test/components/media-player/media-player.test.js
+++ b/test/components/media-player/media-player.test.js
@@ -29,13 +29,7 @@ describe('d2l-labs-media-player', () => {
 				configurable: true,
 				get: () => ({ play: () => Promise.reject(abortError) })
 			});
-			let threw = false;
-			try {
-				await el._play();
-			} catch {
-				threw = true;
-			}
-			expect(threw).to.be.false;
+			await el._play();
 		});
 
 		it('should rethrow non-AbortError errors from play()', async() => {
--- a/test/components/media-player/media-player.test.js
+++ b/test/components/media-player/media-player.test.js
@@ -29,6 +29,8 @@ describe('d2l-labs-media-player', () => {
 				configurable: true,
 				get: () => ({ play: () => Promise.reject(abortError) })
 			});
+			// An unhandled rejection in an async test function causes the test to fail,
+			// so if the AbortError is not suppressed, this call will fail the test
 			await el._play();
 		});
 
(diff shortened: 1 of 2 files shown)


test/components/media-player/media-player.test.js
```diff
@@ -15,4 +15,42 @@ describe('d2l-labs-media-player', () => {
...
+				configurable: true,
+				get: () => ({ play: () => Promise.reject(abortError) })
+			});
+			let threw = false;
+			try {
+				await el._play();
+			} catch {
+				threw = true;
+			}
+			expect(threw).to.be.false;
```
dlockhart: Nit: could use `expect(async() => { await el._play(); }).to.throw(abortError)`

dlockhart (approved):

test/components/media-player/media-player.test.js
rcausey: Oh actually, I'm trying to verify that _no_ error was thrown in this case. I think I could actually simplify this and just call `await el._play()`, since the test will fail if that throws an error.

dlockhart: You can also do `to.not.throw`!

rcausey: Might be missing something, but `expect(async() => { await el._play(); }).to.not.throw()` doesn't actually work here. The test passes, but it also passes if I change the error to `BlahError` - I think because the function is async, it never throws synchronously, so the assertion is always true.

I could use `chai-as-promised` to write `await expect(el._play()).to.not.be.rejected`, but not sure if it's worth adding an extra dependency for this.

`await el._play()` at the end of the test works as expected, since it causes the test to fail if an error is thrown.

dlockhart: Ah ok cool, sorry for sending you astray!

dlockhart (approved):

test/components/media-player/media-player.test.js
rcausey: No worries, thanks for the review!

PR #1013, opened 2026-06-01
ARK-2950 Add Disabled state and Disabled Tooltip to Checkbox Drawer

ChrisLabattD2L:
Context for [ARK-2950](https://desire2learn.atlassian.net/browse/ARK-2950): [Lock CT] d2l-labs-checkbox-drawer - disable changes to the rendered d2l-input-checkbox

As part of a change that we need to make to Completion Tracking, we need to be able to support the ability to disable the `checkbox-drawer` component, fortunately since it's using the [`d2l-input-checkbox`](https://daylight.d2l.dev/components/checkbox-input/#d2l-input-checkbox) component, this is a very change to make!

[ARK-2950]: https://desire2learn.atlassian.net/browse/ARK-2950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -13,6 +13,8 @@ class CheckboxDrawer extends LocalizeLabsElement(SkeletonMixin(LitElement)) {
 		return {
 			checked: { type: Boolean },
 			description: { type: String },
+			disabled: { type: Boolean },
```
dlockhart, line 17: You'll want to default this to `false` in the constructor -- otherwise it's `undefined`... which is "fine" maybe sometimes, but better if it has a nice default.

src/components/checkbox-drawer/checkbox-drawer.js
```diff
@@ -77,6 +79,8 @@ class CheckboxDrawer extends LocalizeLabsElement(SkeletonMixin(LitElement)) {
 				class="d2l-input-checkbox"
 				?skeleton="${this.skeleton}"
 				description=${this.localize(`components:checkboxDrawer:checkbox${this.checked ? 'Expanded' : 'Collapsed'}`)}
+				?disabled="${this.disabled}"
+				disabled-tooltip="${this.disabledTooltip}"
```
dlockhart: Might want to wrap this in `ifDefined(this.disabledTooltip)`.

test/components/checkbox-drawer/checkbox-drawer.vdiff.js
EdwinACL831, line 1: IMO, I think it would worth also to add test for the focus state (for users accessing it using keyboard)

dlockhart: I'm not too worried about that since really this is just wrapping functionality from `<d2l-input-checkbox>`, so this code can trust that focus styles are working correctly.

ChrisLabattD2L: If it's preferred, I can also update my test to use focus instead of hover? Or does it not matter much since it's just wrapping functionality like you said?

dlockhart: I don't think it matters which, what you have is good. For this kind of thing it's less about testing how things look/behave (which is owned by `<d2l-input-checkbox>`) and more just about ensuring that it's wired up correctly.

dlockhart (approved):
Looks good!

PR #1016, opened 2026-06-01
GAUD-9877: optimize the CSS a bit

dlockhart:
@EdwinACL831 here are some proposed changes.

This is a simpler way to do the opacity without needing that `::before` pseudo-element and the custom generated colours. 

From what I can tell, the only downside is that the focus ring gets the 50% opacity as well, which matches the same behaviour as `<d2l-button-subtle>`... seems like an OK trade-off to me.

### DIFF
--- a/src/components/button-ai/button-ai.js
+++ b/src/components/button-ai/button-ai.js
@@ -19,11 +19,6 @@ class ButtonAI extends SlottedIconMixin(ButtonMixin(LitElement)) {
 	static styles = [super.styles, labelStyles, buttonStyles,
 		css`
 			:host {
-				/* comes from --d2l-color-celestine-plus-1 + 0.5 opacity */
-				--d2l-color-celestine-plus-1-opacity-50: #29a6ff80;
-
-				/* comes from --d2l-color-fluorite-plus-1 + 0.5 opacity */
-				--d2l-color-fluorite-plus-1-opacity-50: #ce68fa80;
 				display: inline-block;
 			}
 			:host([hidden]) {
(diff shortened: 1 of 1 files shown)


src/components/button-ai/button-ai.js
```diff
@@ -45,61 +40,30 @@ class ButtonAI extends SlottedIconMixin(ButtonMixin(LitElement)) {
...
-			button:hover:not([disabled]) .content,
-			button:focus:not([disabled]) .content,
-			button:hover:not([disabled]) slot[name="icon"]::slotted(d2l-icon-custom),
-			button:focus:not([disabled]) slot[name="icon"]::slotted(d2l-icon-custom) {
+			:host(:not([disabled])) button:hover,
+			:host(:not([disabled])) button:focus,
+			:host(:not([disabled])) button:hover .property-icon,
+			:host(:not([disabled])) button:focus .property-icon,
+			:host(:not([disabled])) button:hover slot[name="icon"]::slotted(d2l-icon-custom),
+			:host(:not([disabled])) button:focus slot[name="icon"]::slotted(d2l-icon-custom) {
```
dlockhart, line 66: Since `button` doesn't get the `disabled` attribute when using the tooltip, this works in both scenarios.

EdwinACL831:
I see, and you're right that is the behavior with the subtle button. But the regular d2l-button only applies the opacity to the button itself leaving the focus ring as-is, (that was also the request Jeff was doing after confirming with Mike T). So I guess the focus ring in the d2l-button-subtle should also not opaque its focus ring if that makes sense.

dlockhart:
Interesting, yeah I hadn't noticed that `<d2l-button>` was using the same `::before` technique -- I think that's fairly new. It's a pretty simple thing in `<d2l-button>`, but unfortunately it makes the CSS super complex here. 🤔

dlockhart:
Closing as the partially transparent focus ring isn't what we want.

EdwinACL831:
I know, I would say it makes the CSS kind of verbose and sometimes repetitive. But maybe since this is a Labs component, that is not related actually to any of the core components (built it from scratch) this could be a fair enough case to leave these styles. Maybe if (and this is a big if) in the future core supports these ai styles, we can revisit the behavior between the buttons in this cases and its styles across the different states it can have

PR #1050, opened 2026-07-08
GAUD-10267: use modern static properties

dlockhart:
> Disclaimer: this is mostly AI-generated. I've reviewed it myself and do not expect anyone to fully review it line by line.

This is converts from static getters:
```javascript
static get properties() {
  return {
    ...
  };
}
```

To static class level properties:
```javascript
static properties = {
  ...
};
```

And the same for `styles`!

Much easier reviewed with whitespace OFF.

### DIFF
--- a/src/components/opt-in-flyout/opt-out-reason-selector.js
+++ b/src/components/opt-in-flyout/opt-out-reason-selector.js
@@ -5,52 +5,45 @@ import { LocalizeLabsElement } from '../localize-labs-element.js';
 
 class OptOutReasonSelector extends LocalizeLabsElement(LitElement) {
 
-	static get properties() {
-		return {
-			_reasons: { state: true }
-		};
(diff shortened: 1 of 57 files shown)


src/components/opt-in-flyout/opt-out-reason-selector.js
```diff
@@ -5,52 +5,45 @@ import { LocalizeLabsElement } from '../localize-labs-element.js';
...
-	static get properties() {
-		return {
-			_reasons: { state: true }
-		};
-	}
+	static properties = {
+		_reasons: { state: true }
+	};
+
+	static styles = [inputStyles, css`
```
dlockhart, line 12: I made a few minor interventions like this one to be consistent and reduce indentation.

svanherk (approved):

PR #1053, opened 2026-07-10
MED-3088: Expected Behaviour - Athabascau > Athabascau > 20.26.6.22591 > Media Library > Embedded video view transcript option not available when viewing via a mobile device

Syeda-Azim:
### Jira Story
- https://desire2learn.atlassian.net/browse/MED-3088

This change updates the Media Player behavior on iOS so that videos play inline with the Media Player's custom controls available. When a user enters fullscreen, playback transitions to the native iOS video player, where custom controls are no longer available due to platform behavior.

This behavior has been verified in the following areas:

- Content Topics
- Media Library
- Embedded videos (iframe embeds)

https://github.com/user-attachments/assets/8d0a3ced-9d2e-421a-808b-d398db99b210

### DIFF
--- a/src/components/media-player/media-player.js
+++ b/src/components/media-player/media-player.js
@@ -661,7 +661,7 @@ class MediaPlayer extends LocalizeLabsElement(RtlMixin(LitElement)) {
 		};
 		this.afterCaptions = [];
 		this.beforeCaptions = [];
-		this._isoVideoFullscreen = false;
+		this._iosVideoFullscreen = false;
 	}
 
 	get currentTime() {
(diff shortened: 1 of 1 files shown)


src/components/media-player/media-player.js
```diff
@@ -661,6 +661,7 @@ class MediaPlayer extends LocalizeLabsElement(RtlMixin(LitElement)) {
 		};
 		this.afterCaptions = [];
 		this.beforeCaptions = [];
+		this._isoVideoFullscreen = false;
```
MichaelKovalchuk: I think it was supposed to be `this._iosVideoFullscreen` ?

MichaelKovalchuk: what about iOS Audio files? are we supporting similar behavior there too?

Syeda-Azim: Audio files are not affected by this issue because they do not support fullscreen playback. They continue to play within the Media Player, so the custom controls remain available.

MichaelKovalchuk: thanks for confirming

Syeda-Azim: Fixed the variable.

MichaelKovalchuk (approved):

MichaelKovalchuk:
LGTM but it needs to be tested in other browsers to make sure we have no regression

PR #1058, opened 2026-07-14
AUT-5238-step-back-button-for-wizard

SLeAsian:
[AUT-5238](https://desire2learn.atlassian.net/browse/AUT-5238)

Adds Back button support to the Labs wizard component.

Changes include:
- Added **optional** Back button rendering for wizard steps.
- Added demo, vdiff, and unit tests.

Component demo:
https://live.d2l.dev/prs/BrightspaceUI/labs/pr-1058/demo/components/wizard/index.html

[AUT-5238]: https://desire2learn.atlassian.net/browse/AUT-5238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/wizard/wizard-step.js
```diff
@@ -101,6 +120,17 @@ class D2LStep extends LocalizeLabsElement(LitElement) {
...
+	${this.displayBackButton
+		? html`
+			<d2l-button
+				title="${!this.backButtonTooltip ? this.localize('components:wizard:back.button.tooltip') : this.backButtonTooltip}"
+				@click="${this._backClick}"
+			>
+			${!this.backButtonTitle ? this.localize('components:wizard:stepper.defaults.back') : this.backButtonTitle}
+			</d2l-button>`
+		: nothing}
+				</div>
```
NicholasHallman, line 119: It's be a little tidier to have the button html in it's own function, then the ternary becomes

`${this.displayBackButton ? this.#renderBackButton() : nothing}`

src/components/wizard/wizard-step.js
```diff
@@ -119,6 +149,10 @@ class D2LStep extends LocalizeLabsElement(LitElement) {
 		`;
 	}
 
+	_backClick() {
```
NicholasHallman:
```suggestion
	#handleBackClick() {
```

The hash in front makes it truly private. Naming convention is to use "handle" for event handlers

src/components/wizard/wizard.js
```diff
@@ -55,6 +55,14 @@ class D2LWizard extends LitElement {
 		`;
 	}
 
+	back() {
```
NicholasHallman, line 58: Is this suppose to get invoked by the `stepper-back` listener? I don't see where that is connected 🤔

NicholasHallman (commented):
Couple suggestions and a question :)

src/components/wizard/wizard.js
SLeAsian: Yep! It is supposed to be invoked by `stepper-back` listener. It follows the same pattern as `next` and `restart`

NicholasHallman: Oh I see it works the same as next. That pattern seems odd but it's the same as an existing one so I'm not going to suggest we change it, better to keep it the same. 

I think it's odd that the wizard asks the parent to call it's own back function? I would expect the wizard to have all the state it needs to decide if it can go back and if it should. But that change should also need to be done for next and would be a breaking change. So out of scope :)

NicholasHallman (approved):
LGTM thanks for the improvements!

demo/components/wizard/index.html
```diff
@@ -39,11 +39,39 @@ <h2>Basic Wizard</h2>
...
+				<d2l-labs-wizard-step aria-title="This is the second step" restart-button-tooltip="Restart this wizard">
+					<p> second step </p>
+				</d2l-labs-wizard-step>
+
+				<d2l-labs-wizard-step step-title="Almost done" next-button-title="Done" next-button-tooltip="Save this wizard">
+					<p> Last step </p>
+				</d2l-labs-wizard-step>
+			</d2l-labs-wizard>
+			<script>
+				const wizardWithoutBackButton = document.getElementById('wizard-without-back-button');
```
dbatiste: A nice way to avoid the need for the unique `id` values across all the demo snippets is to use:
```javascript
(demo => {
	const wizard = demo.querySelector('d2l-labs-wizard');
	wizard.addEventListener(...);
	 ...
})(document.currentScript.parentNode);
```

src/components/wizard/wizard-step.js
```diff
@@ -22,6 +22,18 @@ class D2LStep extends LocalizeLabsElement(LitElement) {
...
 		},
+		backButtonTitle: {
+			type: String,
+			attribute: 'back-button-title'
+		},
+		backButtonTooltip: {
+			type: String,
+			attribute: 'back-button-tooltip'
+		},
+		displayBackButton: {
```
dbatiste, line 33: Normally we would name this `back-button` since we try to avoid naming properties with verbs (which make them sound like methods). But I see a lot of the existing ones are named like methods already so I guess this is at least consistent. 🤷‍♂️

dbatiste: ex. `hideRestartButton` would have been `restartButtonHidden`, and `disableNextButton` would have been `nextButtonDisabled`.

src/components/wizard/wizard-step.js
```diff
@@ -132,13 +139,52 @@ class D2LStep extends LocalizeLabsElement(LitElement) {
...
+				title="${!this.backButtonTooltip ? this.localize('components:wizard:back.button.tooltip') : this.backButtonTooltip}"
+				@click="${this.#handleBackClick}"
+			>
+			${!this.backButtonTitle ? this.localize('components:wizard:stepper.defaults.back') : this.backButtonTitle}
+			</d2l-button>`;
+	}
+
+	#renderNextButton() {
+		return html`
+			<d2l-button
```
dbatiste, line 165: Nit: normally we'd prefer this formatting, with the attributes ordered:
```html
<d2l-button
	aria-label="${this.nextButtonAriaLabel}"
	class="d2l-labs-wizard-step-button-next"
	@click="${this.#handleNextClick}"
	?disabled="${this.disableNextButton}"
	primary
	title="${!this.nextButtonTooltip ? this.localize('components:wizard:next.button.tooltip') : this.nextButtonTooltip}">
		${!this.nextButtonTitle ? this.localize('components:wizard:stepper.defaults.next') : this.nextButtonTitle}
</d2l-button>
```

test/components/wizard/wizard-step.test.js
```diff
@@ -32,18 +32,51 @@ describe('d2l-labs-wizard-step', () => {
 
 		it('should create stepper-next event', async() => {
 			const elem = await fixture(defaultFixture);
-			elem.addEventListener('stepper-next', (event) => {
-				expect(event.type).to.equal('stepper-next');
-			});
-			elem._nextClick();
+			const eventPromise = oneEvent(elem, 'stepper-next');
```
dbatiste: Nit: this is how we typically assert that events are dispatched:
```javascript
import { clickElem } from '@brightspace-ui/testing';

it('dispatches stepper-next event when next button clicked', async() => {
	const elem = await fixture(defaultFixture);
	clickElem(elem.shadowRoot.querySelector('.d2l-labs-wizard-step-button-next'));
	await oneEvent(elem, 'stepper-next');
});
```
Similar but I think a little cleaner. I don't think asserting the event type is really doing anything, since the test code is setup to only listen for that event.

test/components/wizard/wizard-step.test.js
```diff
@@ -32,18 +32,51 @@ describe('d2l-labs-wizard-step', () => {
...
+			const eventPromise = oneEvent(elem, 'stepper-restart');
+
+			elem.shadowRoot.querySelector('d2l-button').click();
+			const event = await eventPromise;
+
+			expect(event.type).to.equal('stepper-restart');
+		});
+	});
+
+	describe('back button', () => {
```
dbatiste: Are these effectively captured by the visual-diff tests?

src/components/wizard/wizard-step.js
SLeAsian: Since the component is already being used I will leave it for consistency then

test/components/wizard/wizard-step.test.js
SLeAsian: Yeah, the render/no-render behavior is covered by the vdiff tests now. I will drop this test.

dbatiste (approved):

